PR #7, opened 2024-12-27
Add exercises to implement shell tools

illicitonion:
These exercises are language-agnostic - the same description and test files work for any language. In sprint 3 we will set implementing these in JS, and in sprint 4 we will set implementing these in Python.

Sample solutions in https://github.com/CodeYourFuture/Module-Tools/tree/implement-shell-tools-solutions


implement-shell-tools/README.md
```diff
@@ -0,0 +1,9 @@
+# Implement shell tools
+
+Your task is to re-implement shell tools you have used.
```
40thieves, line 3: General thought: do we need to spell it out that the scripts should be run via the `node`/`python` executable? In other words, they won't be making their scripts executable directly?

I would expect some confusion over this since technically they can't run `cat.js` like they can run `cat`

40thieves (approved):
One thought about executables, but generally LGTM! :shipit:

implement-shell-tools/README.md
illicitonion: Good shout - added a clarification

PR #16, opened 2025-02-28
London| Elhadj Abdoul Diallo| Module-Tools | WEEK1 - Individual-shell-tools

eediallo:
## Learners, PR Template

Self checklist

### DIFF
--- a/individual-shell-tools/awk/script-03.sh
+++ b/individual-shell-tools/awk/script-03.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the score from their first attempt.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 1".
+
+awk '{print $1, $3}' scores-table.txt
--- a/individual-shell-tools/awk/script-04.sh
+++ b/individual-shell-tools/awk/script-04.sh
@@ -2,6 +2,6 @@
 
 set -euo pipefail
 
-# TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
-# Your output should contain 3 lines, each with one word and one number on it.
-# The first line should be "Ahmed 4".
+# Output the names of each player in London along with the score from their last attempt.
+# For lines with fewer than 5 columns, print the last column.
+awk '{if (NF >= 5 && $2 == "London") print $1, $5; else if (NF < 5 && $2 == "London") print $1, $NF}' scores-table.txt
--- a/individual-shell-tools/awk/script-07-stretch.sh
+++ b/individual-shell-tools/awk/script-07-stretch.sh
@@ -7,4 +7,3 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total of adding all of that player's scores.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15". The second line should be "Basia 37"
-awk '{ for(i = 3; i <= NF; i++) sum += $i; print $1, sum; sum = 0 }' scores-table.txt
--- a/individual-shell-tools/awk/script-07-stretch.sh
+++ b/individual-shell-tools/awk/script-07-stretch.sh
@@ -7,3 +7,4 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total of adding all of that player's scores.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15". The second line should be "Basia 37"
+awk '{ for(i = 3; i <= NF; i++) sum += $i; print $1, sum; sum = 0 }' scores-table.txt
--- a/individual-shell-tools/cat/script-03.sh
+++ b/individual-shell-tools/cat/script-03.sh
@@ -9,3 +9,4 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+cat -n ../helper-files/helper-3.txt
--- a/individual-shell-tools/cat/script-03.sh
+++ b/individual-shell-tools/cat/script-03.sh
@@ -9,4 +9,4 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
-cat -n ../helper-files/helper-3.txt
+cat ../helper-files/helper-3.txt | nl
--- a/individual-shell-tools/grep/script-01.sh
+++ b/individual-shell-tools/grep/script-01.sh
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt said by the Doctor.
 # The output should contain 6 lines.
+grep ^Doctor dialogue.txt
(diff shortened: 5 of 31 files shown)


individual-shell-tools/grep/script-07.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output, for each `.txt` file in this directory, how many lines of dialogue the Doctor has.
 # The output should show that dialogue.txt contains 6 lines, dialogue-2.txt contains 2, and dialogue-3.txt contains 0.
+grep ^Doctor -icH *.txt
```
mjpeet, line 7: if you need to use the minimum of the options, how would your solution look like? (right now you're using 3)

individual-shell-tools/ls/script-03.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # TODO: Write a command which _recursively_ lists all of the files and folders in this directory _and_ all of the files inside those folders.
 # The output should be a list of names including: child-directory, script-01.sh, helper-1.txt (and more).
 # The formatting of the output doesn't matter.
+cd ../ls
```
mjpeet, line 8: when you consider where this script (script-03.sh) is located, and where it would be executed, do you think this directory change is necessary?

individual-shell-tools/ls/script-04.sh
```diff
@@ -16,8 +16,12 @@ echo "First exercise (sorted newest to oldest):"
 # TODO: Write a command which lists the files in the child-directory directory, one per line, sorted so that the most recently modified file is first.
 # The output should be a list of names in this order, one per line: helper-3.txt, helper-1.txt, helper-2.txt.
 
+cd ../ls/child-directory
+ls -1t
```
mjpeet, line 20: how would you modify this script if using the command cd is not an option? (do you "need" to use cd ?)

individual-shell-tools/sed/script-01.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # TODO: Write a command to output input.txt with all occurrences of the letter `i` replaced with `I`.
 # The output should contain 11 lines.
 # The first line of the output should be: "ThIs Is a sample fIle for experImentIng with sed.".
+sed s/i/I/g input.txt
```
mjpeet, line 8: what would the option -e mean for this script?

individual-shell-tools/sed/script-05.sh
```diff
@@ -6,3 +6,4 @@ set -euo pipefail
 # If a line starts with a number and a space, make the line instead end with a space and the number.
 # So line 6 which currently reads "37 Alisha" should instead read "Alisha 37".
 # The output should contain 11 lines.
+sed -E 's/^([0-9]+) (.*)$/\2 \1/' input.txt
```
mjpeet, line 9: (optional) how would you change this script if you need to use basic regular expressions? (congrats on using ERE ! )

mjpeet (commented):
You're doing great with your scripts! Keep experimenting and practicing. Every script you write improves your skills. Don't hesitate to ask for help and keep up the fantastic work!

individual-shell-tools/awk/script-03.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the score from their first attempt.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 1".
+
+awk '{print $1, $3}' scores-table.txt
```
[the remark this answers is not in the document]
eediallo: That's a good question! I've never tried printf before with ```awk``` but I guess it would allow me to format the result but as the content of the file is alreay in text format print is enough in this case

individual-shell-tools/awk/script-04.sh
```diff
@@ -2,6 +2,6 @@
 
 set -euo pipefail
 
-# TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
-# Your output should contain 3 lines, each with one word and one number on it.
-# The first line should be "Ahmed 4".
+# Output the names of each player in London along with the score from their last attempt.
+# For lines with fewer than 5 columns, print the last column.
+awk '{if (NF >= 5 && $2 == "London") print $1, $5; else if (NF < 5 && $2 == "London") print $1, $NF}' scores-table.txt
```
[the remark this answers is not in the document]
eediallo: That's right!

individual-shell-tools/awk/script-07-stretch.sh
```diff
@@ -7,3 +7,4 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total of adding all of that player's scores.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15". The second line should be "Basia 37"
+awk '{ for(i = 3; i <= NF; i++) sum += $i; print $1, sum; sum = 0 }' scores-table.txt
```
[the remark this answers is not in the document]
eediallo: Yes, I could have done it at the biginning like this: 
```bash
awk '{ sum= 0; for(i = 3; i <= NF; i++) sum += $i; print $1, sum}' scores-table.txt
```

individual-shell-tools/cat/script-03.sh
```diff
@@ -9,3 +9,4 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+cat -n ../helper-files/helper-3.txt
```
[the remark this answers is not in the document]
eediallo: We would have piped it to the ```nl``` command like this: 
```bash
cat  ../helper-files/helper-3.txt | nl
```

individual-shell-tools/grep/script-02.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt that contains the word Doctor (regardless of case).
 # The output should contain 9 lines.
+grep doctor -i dialogue.txt
```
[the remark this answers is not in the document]
eediallo: I agree, going to fix it by following the correct order: ```grep [OPTION...] PATTERNS [FILE...]```

individual-shell-tools/grep/script-01.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt said by the Doctor.
 # The output should contain 6 lines.
+grep ^Doctor dialogue.txt
```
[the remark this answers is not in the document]
eediallo: the result would be the same. The advantage of using the double quote is that it handle speacial characters. This command is working because there is one does not include any special character.

individual-shell-tools/grep/script-07.sh
eediallo: I would remove the -H option and the result will still look the same

individual-shell-tools/ls/script-03.sh
eediallo: It is not neccessary in this case. going to remove it.

individual-shell-tools/ls/script-04.sh
eediallo: I can remove the cd, which I am going to do now.

individual-shell-tools/sed/script-01.sh
eediallo: it specifies multiple editing commands in a single ```sed``` invocation. But is not needed in this case because there is only one command being executed.

mjpeet (commented):
thanks for getting back on the comments!

eediallo:
Thank you @mjpeet for the review and constructive feedback.

illicitonion:

PR #18, opened 2025-02-28
London | Elhadj Abdoul Diallo | Module-Tools | WEEK2- Jq-exercises

eediallo:
## Learners, PR Template

Self checklist

### DIFF
--- a/jq/script-05.sh
+++ b/jq/script-05.sh
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the names of each player, as well as their city.
 # Your output should contain 6 lines, each with two words on it.
+jq -r '.[] | [.name, .city] | join(" ")' scores.json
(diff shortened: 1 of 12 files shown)


jq/script-05.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the names of each player, as well as their city.
 # Your output should contain 6 lines, each with two words on it.
+jq -r '.[] | [.name, .city] | join(" ")' scores.json
```
mjpeet, line 8: most of your solution is based on piping to "join" - this is not a problem at all! but as an example, think about what may be the difference when compared to this solution:

jq -r '.[] | "\(.name) \(.city)"' scores.json

(both solutions have the same result)

mjpeet (commented):
great work, your knowledge in writing scripts to achieve different tasks is getting better!

jq/script-05.sh
eediallo: Thank you! I'll definitely use both of the them.

eediallo:
Thank you for reviewing my work @mjpeet !

illicitonion:

PR #25, opened 2025-03-05
London| Elhadj Abdoul Diallo| Module-Tools | WEEK3 - Implement-shell-tools

eediallo:
## Learners, PR Template

Self checklist


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+    const content = await fs.readFile(path, { encoding: "utf-8" });
+    const lines = extractLinesFromContent(content);
+    printLinesWithOptions(lines);
+  } catch (err) {
+    console.error(err.message);
+  }
+}
+
+function extractLinesFromContent(content) {
+  const lines = content.split("\n");
```
ehwus, line 31: **question:** Why could this line of code make the situation that we've introduced on line 22 worse?

.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+implement-cowsay/.venv.
```
ehwus, line 2: **nitpick:** I think this is leaking in from another assignment, always be careful with which files you're committing.

package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "type": "module",
+    "dependencies": {
+        "commander": "^13.1.0"
```
ehwus, line 4: **praise:** Including this as a dependency increases the portability of the code and made my life easier, thanks!

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
+/**
```
ehwus, line 1: **question:** I like the use of JS doc here, but I would usually expect to see it above a function as a way of typing the parameters and explaining the output. Why did you choose this format over a top level comment explaining the functionality of the script?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
...
+
+program
+  .name("Implement ls")
+  .description("Implements a version of the ls command")
+  .option("-1, --one", "Output all files and directories each in a line")
+  .option("-a, --hidden", "Output hidden files/directories")
+  .argument("[paths...]", "the paths to be processed")
+  .parse(process.argv);
+
+const filePaths = program.args.length ? program.args : ["."];
```
ehwus, line 21: **praise:** This is a good way of implementing a sensible default, and shows that you know a great deal about how `ls` works

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
...
+
+const one = program.opts().one;
```
ehwus, line 23: **question:** could we name these something that makes the code easier to read?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
...
+
+async function listDirectoryContents(filePath) {
+  try {
+    const files = await fs.readdir(filePath, { withFileTypes: true }); // is returned as a Dirent
+    const filteredFiles = files
+      .filter((file) => hidden || !file.name.startsWith("."))
+      .map((file) => file.name);
+
+    if (hidden) {
+      filteredFiles.unshift(".", "..");
```
ehwus, line 34: **praise:** This is a smart way of replicating the showing of directory functionality within ls

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
...
+
+const filePaths = program.args.length ? program.args : ["."];
+
+const one = program.opts().one;
+const hidden = program.opts().hidden;
+
+async function listDirectoryContents(filePath) {
+  try {
+    const files = await fs.readdir(filePath, { withFileTypes: true }); // is returned as a Dirent
+    const filteredFiles = files
```
ehwus, line 29: **question:** could we avoid iterating over this multiple times here?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
...
+      filteredFiles.forEach((file) => console.log(file));
+    } else {
+      console.log(filteredFiles.join(" "));
+    }
+  } catch (err) {
+    console.error(`Error reading directory ${filePath}:`, err.message);
+  }
+}
+
+filePaths.map(listDirectoryContents);
```
ehwus, line 47: **question:** Why have you chosen to use a map here?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Aggregates numerical data from an array of file content strings.
+ *
+ * @param {string[]} files - An array of strings, each representing the content of a file.
+ * Each string should contain space-separated numbers.
+ * @returns {number[]|number} - An array of sums for each column of numbers if there are multiple files,
+ * or 0 if there is only one file.
+ */
+function aggregateFileData(fileCounts) {
```
ehwus, line 90: **question:** Please could you talk me through this function line by line?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,103 @@
...
+function removeEndEmptyLine(arr) {
+  return arr[arr.length - 1] === "" ? arr.slice(0, -1) : arr;
+}
+
+const lineOption = program.opts().line;
+const charOption = program.opts().char;
+const wordOption = program.opts().word;
+
+async function countLinesWordsCharsInFile(path) {
+  const content = await fs.readFile(path, { encoding: "utf-8" });
```
ehwus, line 25: **question:** What are the tradeoffs of reading files in this way?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,103 @@
...
+
+  const lines = removeEndEmptyLine(content.split("\n"));
```
ehwus, line 27: **question:** These are clever ways of getting the counts, but do we always need to do all of this work?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,103 @@
...
+    element.split(" ").filter((word) => word !== "")
+  );
+
+  const chars = content.split("");
+
+  const numberOfLines = lines.length;
+  const numberOfWords = words.length;
+  const numberOFChars = chars.length;
+
+  if (lineOption) {
```
ehwus, line 39: **question:** What if I'd like to have the lines and the chars?

Screenshot 2025-03-19 at 08 46 07

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,103 @@
...
+
+  if (wordOption) {
+    return `${numberOfWords} ${path}`;
+  }
+
+  return `${numberOfLines} ${numberOfWords} ${numberOFChars} ${path}`;
+}
+
+async function processFilesAndDisplayCounts() {
+  const fileCounts = await Promise.all(args.map(countLinesWordsCharsInFile));
```
ehwus, line 55: **praise:** awaiting and using `Promise.all` with a map here makes a lot of sense and makes sure we can process as many files as are needed, great stuff

ehwus (commented):
**cat**

You've produced a readable and working replica of the `cat` shell tool here, and it shows a solid understanding of node fundamentals and using `commander` to build scripts. There are a few tweaks that we could make to get this to be as efficient as cat is, and I would focus on:

- Learning about operating on files in a memory efficient manner
- Why and when to await the response of promises
- JavaScript array iteration - which to use when and why

Overall though this is a really solid effort and it was great to see you reach such parity with the `cat` tool with your own code.

**ls**

Once again this is a good implementation of `ls` and the code is clear and readable. There are a few flags for efficiency and clarity, but overall a very good effort.

**wc**

Once again this largely replicates `wc` well barring a few little logic tweaks that I have raised, but parts of the code feel complicated to follow and we have also fallen into some of the previously discussed patterns surrounding reading from the filesystem. It feels like it doesn't have the readability of the other two scripts, so I'm curious what the difference was when you wrote it - did you find it a harder problem to solve?

**Overall**

This is a great demonstration of your knowledge of shell toolings and how to replicate this functionality well in Node, which you were able to do successfully. Keep tradeoffs in mind, particularly when working with the file system, and make sure that you have a good understanding of when to use each array function and when it's useful to await promises.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+
+program
+  .name("Implement cat")
+  .description("Implement a version of the cat program")
+  .option("-n, --number", "Number the output lines, starting at 1")
+  .option("-b, --number2", "Number the nonempty lines, starting at 1")
+  .argument("<paths...>", "The file paths to process")
+  .parse(process.argv);
+
+let args = program.args;
```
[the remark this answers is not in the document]
eediallo: we are not. It makes sense to use the `const` keyword here.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+//console.log(args)
```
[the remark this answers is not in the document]
eediallo: ok. going to remove it. Thanks

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+
+const nOption = program.opts().number;
```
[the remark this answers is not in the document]
eediallo: will do. Thanks. I was actually struggling to find meaning names for these options.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+const bOption = program.opts().number2;
+let lineNumber = 1;
+
+async function readAndPrintFileContent(path) {
+  try {
+    const content = await fs.readFile(path, { encoding: "utf-8" });
```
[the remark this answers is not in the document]
eediallo: It's reading the content of the file(path) provided through the CLI asyncronously which means it won't prevent other tasks from executing. Could we do this more memory efficiently? I am not 100% sure but I know if I was reading a large file, I would have used `Stream` to read it in chuncks instead of loading it once.

implement-shell-tools/cat/cat.js
eediallo: I am not sure. Could you please advice what would be the best practice here?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+  if (lines[lines.length - 1] === "") {
+    lines.pop(); // excludes last line if empty
+  }
+  return lines;
+}
+
+function printLinesWithOptions(lines) {
+  lines.forEach((line) => {
+    if (nOption) {
+      console.log(`${lineNumber++} ${line}`);
```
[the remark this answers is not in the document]
eediallo: It is possible by using the teneray operator

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+        console.log(`${lineNumber++} ${line}`);
+        return;
+      }
+    }
+
+    console.log(line);
+  });
+}
+// display file/s contents with line numbers.
+await Promise.all(args.map(readAndPrintFileContent));
```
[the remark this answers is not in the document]
eediallo: That's a good observation. Since the `readAndPrintFileContent` function is not returning a promise here, there is no need to use both `await` and `Promise.all`. We are using `map` here because we may get multiple files to read from the command line.

implement-shell-tools/cat/cat.js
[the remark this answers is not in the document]
ehwus: You're along the right lines here, text files might be absolutely massive (e.g. logs) so making sure that our programs operate on them efficiently is really important, especially as we're processing line by line anyway

implement-shell-tools/cat/cat.js
ehwus: If we've loaded the entire file into memory on line 22, what are we doing here?

implement-shell-tools/ls/ls.js
eediallo: No reason to be honest. I agree that it makes more sense to put above the function. I am going to do just that.

implement-shell-tools/ls/ls.js
eediallo: Yes. I am going to update it.

implement-shell-tools/cat/cat.js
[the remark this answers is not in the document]
ehwus: > There are only two hard things in Computer Science: cache invalidation and naming things.
> 
>  -- Phil Karlton

implement-shell-tools/ls/ls.js
eediallo: because `filePaths is an array`. I could have used a `for loop`` instead. This is to make sure every filePath is processed in the directory

implement-shell-tools/wc/wc.js
eediallo: if the file was large, that would caused high memory consumption and possible out of memory errors.

implement-shell-tools/ls/ls.js
ehwus: What is the benefit of `map` vs. `forEach`

implement-shell-tools/wc/wc.js
eediallo: maybe not. may I could have write the code instead of creating a separate function for it.

implement-shell-tools/wc/wc.js
eediallo: Right. I've not implemented this one. I am going to  add the functionality

implement-shell-tools/cat/cat.js
eediallo: we are processing it once.

implement-shell-tools/ls/ls.js
eediallo: do something multiple times.

eediallo: That's a good question. I need to search that. Personally i use`map` because it allows me to write less code. But I am going to google this to find out.

implement-shell-tools/wc/wc.js
eediallo: Sure  the `aggregatedFilesData ` is an array with three possible values: sum of lines, chars or words. The following code: 
```bash
 if (aggregatedFilesData !== 0 && lineOption && !(wordOption || charOption)) {
    console.log(`${aggregatedFilesData[0]} total`);
    return;
  }
``` 
whether aggregatedFilesData is not 0 and line option is provided but not the rest of the options(chars and words).if it is true, it means only sum of lines is available on `aggregatedFilesData` array therefore the sum is displayed with `total` in front of it.

it is the same for chars option in the following code: 
```bash
if (aggregatedFilesData !== 0 && charOption && !(wordOption || lineOption)) {
    console.log(`${aggregatedFilesData[0]} total`);
    return;
  }
```

Here: 
```bash
if (
    aggregatedFilesData !== 0 &&
    ((lineOption && charOption) || (lineOption && wordOption))
  ) {
    console.log(`${aggregatedFilesData[0]} ${aggregatedFilesData[1]} total`);
    return;
  }
```
This code checks if two options are provided, it could be line option and char option or line and word option. if it is true, I know the `aggregatedFilesData` array contains two values which are then displayed alongside the `total` string. 

The `return` keyword in each statement ensure that the execution does not continue.

if the bove conditions are false: the sum of each option is display if `aggregatedFilesData` is not `0`, hence the code below: aggregatedFilesData !== 0
    ? console.log(
        `${aggregatedFilesData[0]} ${aggregatedFilesData[1]} ${aggregatedFilesData[2]} total`
      )
    : null;

implement-shell-tools/cat/cat.js
ehwus: We're copying the entirety of the file into memory again in an array, so we've doubled our problem

eediallo: OK. I see. Googling to see how to avoid this.

eediallo: I have refactored it by removing the `extractLinesContent` function and extracting lines directly in `readAndPrintFileContent` function. Would this solve the problem?

implement-shell-tools/wc/wc.js
eediallo: I refactor this code too to avoid copying the entire array inot memory again.

illicitonion:

PR #27, opened 2025-03-05
London | DONARA BLANC | Module Tools | SPRINT 1 | Individual-Shell-Tools

donarbl:
Self checklist

- [ X] I have committed my files one by one, on purpose, and for a reason
- [ X] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ X] I have tested my changes
- [X ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ X] My changes meet the [requirements](./README.md) of this task

## Changelist

Testing shell tools to extract information from scripts.

## Questions

### DIFF
--- a/individual-shell-tools/awk/script-04.sh
+++ b/individual-shell-tools/awk/script-04.sh
@@ -5,3 +5,8 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
 # Your output should contain 3 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4".
+
+awk '/London/ { if (NF<5) print $1, $NF; else print $1, $5}' scores-table.txt
+Ahmed 4
+Basia 6
+Leila 1
--- a/individual-shell-tools/cat/script-03.sh
+++ b/individual-shell-tools/cat/script-03.sh
@@ -9,3 +9,5 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+
+cat ../helper-files/helper-3.txt | nl # ordered by numbers 
--- a/individual-shell-tools/grep/script-07.sh
+++ b/individual-shell-tools/grep/script-07.sh
@@ -4,3 +4,16 @@ set -euo pipefail
 
 # TODO: Write a command to output, for each `.txt` file in this directory, how many lines of dialogue the Doctor has.
 # The output should show that dialogue.txt contains 6 lines, dialogue-2.txt contains 2, and dialogue-3.txt contains 0.
+
+grep ^Doctor -c *.txt # this command gaves the expected output
+
+dialogue.txt:6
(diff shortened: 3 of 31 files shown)


individual-shell-tools/awk/script-04.sh
```diff
@@ -5,3 +5,8 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
 # Your output should contain 3 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4".
+
+awk '/London/ { if (NF<5) print $1, $NF; else print $1, $5}' scores-table.txt
```
ehwus, line 9: **praise:** Good use of conditional logic in the awk command here!

individual-shell-tools/cat/script-03.sh
```diff
@@ -9,3 +9,5 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+
+cat ../helper-files/helper-3.txt | nl # ordered by numbers
```
ehwus, line 13: **question:** This is good and works, but I'm curious if we can do it without the pipe?

individual-shell-tools/grep/script-07.sh
```diff
@@ -4,3 +4,16 @@ set -euo pipefail
...
 # TODO: Write a command to output, for each `.txt` file in this directory, how many lines of dialogue the Doctor has.
 # The output should show that dialogue.txt contains 6 lines, dialogue-2.txt contains 2, and dialogue-3.txt contains 0.
+
+grep ^Doctor -c *.txt # this command gaves the expected output
+
+dialogue.txt:6
+dialogue-2.txt:2
+dialogue-3.txt:0
+
+grep ^ Doctor -c *.txt # however here i put a space between the carrot and the search name and it gave me a different output, and i dont know why
```
ehwus, line 14: **question:** How could we find out what went wrong with that command?

ehwus (commented):
This is a great submission Donara - it shows that you have grasped the key concepts introduced so far and a great foundational knowledge of coding. There are a few comments to address around how to present bash scripts in a way that are executable and reviewable.

I can see that this felt quite difficult at times, and I'm grateful that you left comments that let me know your thought process and struggles here. I'd just like to assure you that this is a normal part of the process, you're learning things that are complex and antiquated, and that work is paying off. I would like to suggest though (and feel free to ignore this) that you approach these problems with a curiosity - there is an inevitability to the feeling of not immediately knowing the answer, that is a part of the job. If you can approach these challenges with a growth mindset, resigning yourself to struggle being a part of the process of computing, and following the problems through as best you can to a solution (even if that means asking for help rather than submitting a couple of problems that are unsolved), I think you'd be better for it.

You're making a really good start to something that is very challenging and please keep it up and keep the faith!

PR #39, opened 2025-03-08
London | Elhadj Abdoul Diallo| Module-Tools | WEEK4 - Implement-cowsay

eediallo:
## Learners, PR Template

Self checklist


implement-cowsay/cow.py
```diff
@@ -0,0 +1,14 @@
+import cowsay # type: ignore
```
illicitonion, line 1: What does `# type: ignore` mean here? Why did you add it?

implement-cowsay/cow.py
```diff
@@ -0,0 +1,14 @@
...
+import argparse
+
+parser = argparse.ArgumentParser(description="Make animals say things")
+parser.add_argument('--animal', choices=cowsay.char_names, help="The animal to be saying things", default="cow")
```
illicitonion, line 5: It's not a problem, but is a bit unusual to have mixed quotes (some of the strings on this line are wrapped in `'`s and others `"`s) - is there a reason to prefer one over the other? Or a reason you used different ones here?

illicitonion (commented):
Looking good - I left a few questions to ponder :)

implement-cowsay/cow.py
eediallo: I added this because the editor was complaining about it thinking it is a built-in package. But in reality it is installed through the `venv` and is available in the my `requirements.txt` file. It does not have any effect here.

implement-cowsay/cow.py
eediallo: No reason to be honest. Python accept both. I know I should pick one and be consistent with it.

illicitonion: Makes sense :) Personally I prefer `"`s because it's more often that I write a string containing a `'` than a `"`, e.g. I will often write a string literal like `"I don't know"` but I less often write a string literal like `'I said "hello"'`

implement-cowsay/cow.py
illicitonion: Makes sense - thanks!

illicitonion:

PR #42, opened 2025-03-11
London | Samira Hekmati | Module Tools | Week 2 | jq

samirahekmati:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/jq/script-05.sh
+++ b/jq/script-05.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the names of each player, as well as their city.
 # Your output should contain 6 lines, each with two words on it.
+
+jq -r '.[] | "\(.name) \(.city)"' scores.json
--- a/jq/script-07.sh
+++ b/jq/script-07.sh
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the score from their last attempt.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4" with no quotes.
+
+jq -r '.[] | .scores[-1] as $scores | "\(.name) \($scores)" ' scores.json
--- a/jq/script-07.sh
+++ b/jq/script-07.sh
@@ -7,4 +7,4 @@ set -euo pipefail
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4" with no quotes.
 
-jq -r '.[] | .scores[-1] as $scores | "\(.name) \($scores)" ' scores.json
+jq -r '.[] | .scores[-1] as $lastScore | "\(.name) \($lastScore)" ' scores.json
--- a/jq/script-09.sh
+++ b/jq/script-09.sh
@@ -7,4 +7,12 @@ set -euo pipefail
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15" with no quotes.
 
+echo "first solution"
+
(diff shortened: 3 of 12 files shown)


jq/script-05.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the names of each player, as well as their city.
 # Your output should contain 6 lines, each with two words on it.
+
+jq -r '.[] | "\(.name) \(.city)"' scores.json
```
illicitonion, line 9: In script-03 you used `join` and here you used string interpolation - both work, and both could be used in both places - do you have any thoughts on which you prefer / which is more clear?

jq/script-07.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the score from their last attempt.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4" with no quotes.
+
+jq -r '.[] | .scores[-1] as $scores | "\(.name) \($scores)" ' scores.json
```
illicitonion, line 10: `$scores` seems like a bit of a misleading variable name here, can you think why? What could be more clear?

jq/script-09.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total scores from all of their games added together.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15" with no quotes.
+
+jq -r '.[] | "\(.name) \([.scores[]] | add)"' scores.json
```
illicitonion, line 10: What does `[.scores[]]` here do? Could you write this more simply?

illicitonion (commented):
Looking good! This all works well, but I left a few questions about some specifics :)

jq/script-05.sh
samirahekmati: **Thank you so much Daniel, @illicitonion for your feedback!** I agree that both methods join(", ") and string interpolation are valid, and it's interesting to think about their usage in different contexts.

For example in script-03, I chose join(", ") as I felt it was a **cleaner way to handle the array of values** (in this case, name and profession). It also keeps the structure clear by dealing with the array directly. On the other hand, in script-05, I used string interpolation **because it was more intuitive for handling the dynamic insertion of values within a string**.

I don't have a strong preference for one over the other, but I can see how string interpolation may be more flexible, especially if we were working with more complex strings or needing more control over formatting.

I think, if clarity and simplicity are the primary goals, join(", ") works well for array type of structures, while string interpolation might be more readable when we're dealing with individual elements or values.

jq/script-07.sh
samirahekmati: I get what you're trying to say. The variable `scores` can be misleading because it implies it holds all the scores, but it actually holds just the last score. Renaming it to something like
`lastScore` would make it clearer that it's referring to the most recent score, not the entire list.

jq/script-09.sh
samirahekmati: The part `[.scores[]]`  is iterating over each score for every player, then the add function is summing those scores together. 
`.scores[]` is a shorthand way to work with each element inside the scores array. It pulls out each score and feeds it into the add function, which then sums them up. It works, but it's a bit more complicated than needed.
To simplify it, I can directly sum the scores by using the add function on the .scores array without needing the extra []. 

`jq -r '.[] | "\(.name) \((.scores | add))"' scores.json`

illicitonion:
Perfect - thanks so much!

illicitonion:

PR #43, opened 2025-03-11
LONDON | DONARA BLANC | MODULE TOOLS | SPRINT 2| SHELL PIPELINES

donarbl:
Self checklist

- [x ] I have committed my files one by one, on purpose, and for a reason
- [x ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [x ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ x] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/shell-pipelines/ls-grep/script-03.sh
+++ b/shell-pipelines/ls-grep/script-03.sh
@@ -4,3 +4,11 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+ls sample-files | grep -E '^[A-Z][^A-Z]*$'
+Ariane
+Daniel
+Levi
+London
+Niamh
+Olga
+York
--- a/shell-pipelines/sort-uniq-head-tail/script-02.sh
+++ b/shell-pipelines/sort-uniq-head-tail/script-02.sh
@@ -5,3 +5,11 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
 # The first line of your output should be "Basia London 22 9 6" (with no quotes).
+sort -nrk3 scores-table.txt
+
+Basia London 22 9 6
+Piotr Glasgow 15 2 25 11 8
+Chandra Birmingham 12 6
+Mehmet Birmingham 3 12 17
+Leila London 1
+Ahmed London 1 10 4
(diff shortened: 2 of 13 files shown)


shell-pipelines/ls-grep/script-03.sh
```diff
@@ -4,3 +4,11 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+ls sample-files | grep -E '^[A-Z][^A-Z]*$'
```
illicitonion, line 7: What does the `-E` do here? Why do you need it here when you didn't need it for script-02 which otherwise looks quite similar?

shell-pipelines/sort-uniq-head-tail/script-02.sh
```diff
@@ -5,3 +5,11 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
 # The first line of your output should be "Basia London 22 9 6" (with no quotes).
+sort -nrk3 scores-table.txt
```
illicitonion, line 8: This works, but personally I tend to separate out flags which take values to make it more clear which values are associated with which flag, i.e. I would probably write `sort -nr -k3 scores-table.txt` or `sort -n -r -k3 scores-table.txt` just to make it a little more clear the 3 is attached to the `-k` :)

shell-pipelines/sort-uniq-head-tail/script-07.sh
```diff
@@ -6,3 +6,4 @@ set -euo pipefail
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
 # The word "Event" should not appear in your script's output.
+tail -9 events-with-timestamps.txt | sort -k3 | awk '{print $3}' | uniq -c
```
illicitonion, line 9: This works, but I probably would've started it with `head -n+2` rather than `tail -9` - can you think about why that is? What's the difference between the two?

shell-pipelines/tr/script-02.sh
```diff
@@ -7,3 +7,14 @@ set -euo pipefail
 # so every Y should be a Z, and every Z should be a Y!
 #
 # TODO: Write a command to output the contents of text.txt with every Y and Z swapped (both upper and lower case).
+Dear Zara,
```
illicitonion, line 10: I think the command is missing here? Can you add it?

illicitonion (commented):
Looking good! I left a few comments - also the general shell script comments from https://github.com/CodeYourFuture/Module-Tools/pull/44 apply here too :)

PR #44, opened 2025-03-11
LONDON | DONARA BLANC | MODULE TOOLS | SPRINT 2 | JQ exercises

donarbl:
Self checklist

- [x ] I have committed my files one by one, on purpose, and for a reason
- [ x] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ x] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ x] My changes meet the [requirements](./README.md) of this task

## Changelist

Training on JQ commands and flags and pipes and chaining them together to extract information.

## Questions

### DIFF
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+jq -r .'name' person.json
+Selma
--- a/jq/script-03.sh
+++ b/jq/script-03.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person, then a comma, then their profession.
 # Your output should be exactly the string "Selma, Software Engineer", but should not contain any quote characters.
+q -r '[.name, .profession] | join (",")' person.json
+Selma,Software Engineer
--- a/jq/script-05.sh
+++ b/jq/script-05.sh
@@ -5,3 +5,11 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the names of each player, as well as their city.
 # Your output should contain 6 lines, each with two words on it.
+jq -r '.[] | "\(.nam
+e) \(.city)"' scores.json
+Ahmed London
+Basia London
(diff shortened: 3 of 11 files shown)


jq/script-01.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+jq -r .'name' person.json
+Selma
```
illicitonion, line 9: It is useful to see the output produced, but we'd generally avoid putting this directly in a shell script, as it causes the script to no longer run, e.g.

```console
% ./script-01.sh
Selma
./script-01.sh: line 9: Selma: command not found
```

If you're going to include the output, I'd recommend doing so in a comment (i.e. putting a `#` before the output) so that the scripts can still run :)

jq/script-03.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person, then a comma, then their profession.
 # Your output should be exactly the string "Selma, Software Engineer", but should not contain any quote characters.
+q -r '[.name, .profession] | join (",")' person.json
```
illicitonion, line 8: It looks like something went slightly wrong in your copy+paste here?

But this raises an interesting workflow question - there are two ways you can go about making this kind of script - you can work in the terminal then copy+paste into the file, or work in the file and run the file each time you want to try something out. How do you feel about the two? What are the advantages/disadvantages of each way of working?

jq/script-05.sh
```diff
@@ -5,3 +5,11 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the names of each player, as well as their city.
 # Your output should contain 6 lines, each with two words on it.
+jq -r '.[] | "\(.nam
+e) \(.city)"' scores.json
```
illicitonion, line 9: How did you end up with a newline in the middle of your command here? (Same for script-06 and script-07, script-10, and script-11)?

illicitonion (commented):
These are all looking good, congrats!

I left a few comments about working with shell scripts in general, but all the jq stuff here is looking great :)

PR #49, opened 2025-03-20
London | Ameneh Keshavarz |  Implement-shell-tools| WEEK3

Ameneh-Keshavarz:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -11,8 +11,9 @@ program
   .parse(process.argv);
 
 let args = program.args;
-const nOption = program.opts().number;
-const bOption = program.opts().number2;
+//const nOption = program.opts().number;
+//const bOption = program.opts().number2;
+const { number: nOption, number2: bOption } = program.opts();
 let lineNumber = 1;
 let nonEmptyLineNumber = 1;
 
@@ -28,10 +29,13 @@ function printLinesWithOptions(lines) {
   });
 }
 
-async function readAndPrintFileContent(path) {
+async function readFileContent(path) {
   try {
     const content = await fs.readFile(path, { encoding: "utf-8" });
     const lines = content.split("\n");
+    if(lines[lines.length - 1] === "") {
+      lines.pop();
+    }
     printLinesWithOptions(lines);
   } catch (err) {
     console.error(`Error reading file ${path}: ${err.message}`);
@@ -39,5 +43,5 @@ async function readAndPrintFileContent(path) {
 }
 
 args.forEach((arg) => {
-  readAndPrintFileContent(arg);
+  readFileContent(arg);
 });
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -18,9 +18,9 @@ let nonEmptyLineNumber = 1;
 function printLinesWithOptions(lines) {
   lines.forEach(line => {
     if (nOption) {
-      console.log(`${lineNumber++} ${line}`);
+      console.log(`${String(lineNumber++).padStart(6)}  ${line}`);
     } else if (bOption && line.trim()) {
-      console.log(`${nonEmptyLineNumber++} ${line}`);
+      console.log(`${String(nonEmptyLineNumber++).padStart(6)}  ${line}`);
     } else {
       console.log(line);
     }
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -11,8 +11,7 @@ program
   .parse(process.argv);
 
 let args = program.args;
-//const nOption = program.opts().number;
-//const bOption = program.opts().number2;
+
 const { number: nOption, number2: bOption } = program.opts();
 let lineNumber = 1;
 let nonEmptyLineNumber = 1;
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,0 +1,43 @@
+const { promises: fs } = require("fs");
+const process = require("process");
+const { program } = require("commander");
+
+program
+  .name("cat")
+  .description("Implement a version of the cat program")
+  .option("-n, --number", "Number the output lines, starting at 1")
+  .option("-b, --number2", "Number the nonempty lines, starting at 1")
+  .argument("<paths...>", "The file paths to process")
+  .parse(process.argv);
+
+let args = program.args;
+const nOption = program.opts().number;
+const bOption = program.opts().number2;
+let lineNumber = 1;
+let nonEmptyLineNumber = 1;
+
+function printLinesWithOptions(lines) {
+  lines.forEach((line, index) => {
+    if (nOption) {
+      console.log(`${lineNumber++} ${line}`);
+    } else if (bOption && line.trim()) {
+      console.log(`${nonEmptyLineNumber++} ${line}`);
+    } else {
+      console.log(line);
+    }
+  });
+}
+
+async function readAndPrintFileContent(path) {
+  try {
+    const content = await fs.readFile(path, { encoding: "utf-8" });
+    const lines = content.split("\n");
+    printLinesWithOptions(lines);
+  } catch (err) {
+    console.error(`Error reading file ${path}: ${err.message}`);
+  }
+}
+
+args.forEach((arg) => {
+  readAndPrintFileContent(arg);
+});
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,26 +1,25 @@
 const { promises: fs } = require("fs");
-const process = require("process");
 const { program } = require("commander");
 
 program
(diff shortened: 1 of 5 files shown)


implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,71 @@
+const { promises: fs } = require('fs');
+const path = require('path');
+const { program } = require('commander');
+
+program
+  .description('Count lines, words, and characters in the specified files')
+  .option('-l, --lines', 'Count the lines')
```
blorente, line 7: When counting lines, your program counts one more than `wc`:

```shell
➜  wc -l sample-files/3.txt
       5 sample-files/3.txt

➜ node wc.js -l sample-files/3.txt
6 /Users/blorente/code/github.com/CodeYourFuture/Module-Tools/implement-shell-tools/wc/sample-files/3.txt
```

Why does this happen? How could we fix it so that we print the right number of lines?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,71 @@
...
+  console.error("Error: No files provided.");
+  process.exit(1);
+}
+
+async function countFileStats(filePath, options) {
+  try {
+    const data = await fs.readFile(filePath, 'utf8');
+    
+    const lines = data.split('\n').length;
+    const words = data.split(/\s+/).filter(Boolean).length;
```
blorente, line 25: This section of the code is processing `data` twice: Once to split it into lines, and another one to split it into words. While this is correct, in very large files, this could potentially be quite slow. 

Is there a way we could avoid processing the entirety of `data` twice?

This is a stretch goal, as the current program is already correct.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,71 @@
...
+      totalWords += result.words;
+      totalCharacters += result.characters;
+    }
+  }
+
+  if (files.length > 1) {
+    const totals = [];
+    if (options.lines !== false) totals.push(totalLines);
+    if (options.words !== false) totals.push(totalWords);
+    if (options.characters !== false) totals.push(totalCharacters);
```
blorente, line 63: Why did you add the `!== false` clause to these lines? Is there any way we could express the same in a more concise way?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,71 @@
+const { promises: fs } = require('fs');
+const path = require('path');
+const { program } = require('commander');
+
+program
+  .description('Count lines, words, and characters in the specified files')
+  .option('-l, --lines', 'Count the lines')
+  .option('-w, --words', 'Count the words')
+  .option('-c, --characters', 'Count the characters')
+  .parse(process.argv);
```
blorente, line 10: Your solution prints the full path of each file: 

```shell
➜ node wc.js -l sample-files/3.txt
6 /Users/blorente/code/github.com/CodeYourFuture/Module-Tools/implement-shell-tools/wc/sample-files/3.txt
```

Whereas `wc` only prints the path relative to the directory we run it from:
```shell
➜  wc -l sample-files/3.txt
       5 sample-files/3.txt
```

How would you modify this code to print the relative path?

blorente (commented):
Thank you for your PR! I'm Borja, a volunteer, and I'll be reviewing your code.

The implementations of `cat`, `ls` and `wc` are good! I have left a few edge cases that are not covered, as well as a few suggestions, but you're definitely on the right track :)

Besides that, would you mind including the `package.json` and `package-lock.json` files you used during development? For a stretch goal, could you tell me why we should or shouldn't commit them to a repository?

Keep up the good work!

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,47 @@
...
+const { number: nOption, number2: bOption } = program.opts();
+let lineNumber = 1;
+let nonEmptyLineNumber = 1;
+
+function printLinesWithOptions(lines) {
+  lines.forEach((line, index) => {
+    if (nOption) {
+      console.log(`${lineNumber++} ${line}`);
+    } else if (bOption && line.trim()) {
+      console.log(`${nonEmptyLineNumber++} ${line}`);
```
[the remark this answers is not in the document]
Ameneh-Keshavarz: Thanks for the feedback. I checked my code, and the issue was that I expected line numbers to reset for each file. But the real cat -b also keeps numbering across multiple files without resetting. I've updated my script, and it now matches that behavior.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,47 @@
...
+//const nOption = program.opts().number;
+//const bOption = program.opts().number2;
+const { number: nOption, number2: bOption } = program.opts();
+let lineNumber = 1;
+let nonEmptyLineNumber = 1;
+
+function printLinesWithOptions(lines) {
+  lines.forEach((line, index) => {
+    if (nOption) {
+      console.log(`${lineNumber++} ${line}`);
```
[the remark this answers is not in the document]
Ameneh-Keshavarz: Yes, line numbers were misaligned. Fixed it using padStart() to align them like cat -n.

implement-shell-tools/wc/wc.js
Ameneh-Keshavarz: I was using data.split('\n').length, which overcounts lines. To match the behavior of wc -l, I updated it to:     
const lines = (data.match(/\n/g) || []).length;

implement-shell-tools/wc/wc.js
Ameneh-Keshavarz: The !== false checks were overly cautious, siince commander sets options to true or undefined, the concise and correct way is simply:
if (options.lines)

implement-shell-tools/wc/wc.js
Ameneh-Keshavarz: I was printing the absolute path using path.resolve, which made the output differ from wc. I’ve updated the code to print the original file argument instead, so it now shows the relative path the user typed.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,47 @@
...
+      console.log(line);
+    }
+  });
+}
+
+async function readFileContent(path) {
+  try {
+    const content = await fs.readFile(path, { encoding: "utf-8" });
+    const lines = content.split("\n");
+    if(lines[lines.length - 1] === "") {
```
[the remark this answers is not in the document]
Ameneh-Keshavarz: Thanks, the special case isn’t strictly necessary, but I included it to be safe in case the input source changes in the future. I’ll also make sure to handle potential empty results from split() to avoid issues as the code evolves.

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,37 @@
+const fs = require('fs');
+const { program } = require('commander');
+
+program
+  .option('-a, --all', 'Show hidden files')
+  .option('-1', 'Show one file per line')
+  .argument('[dirPath]', 'Directory path', '.')
+  .parse(process.argv);
```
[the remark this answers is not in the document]
Ameneh-Keshavarz: Thanks, I removed the unnecessary special case for single input and added support for multiple directory arguments like node ls.js dir1 dir2. I also made sure the code handles each path safely and gives clear output.

Ameneh-Keshavarz:
Hi Borja,

Thanks for the feedback and for reviewing my code! I looked at the edge cases you mentioned and add the package.json and package-lock.json files.

We should commit both files because: package.json shows what packages the project needs. package-lock.json makes sure everyone installs the exact same versions. Thanks again

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,47 @@
...
+  .name("cat")
+  .description("Implement a version of the cat program")
+  .option("-n, --number", "Number the output lines, starting at 1")
+  .option("-b, --number2", "Number the nonempty lines, starting at 1")
+  .argument("<paths...>", "The file paths to process")
+  .parse(process.argv);
+
+let args = program.args;
+//const nOption = program.opts().number;
+//const bOption = program.opts().number2;
```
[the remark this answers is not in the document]
Ameneh-Keshavarz: Thanks, I removed the unnecessary commnets.

blorente (commented):
Hi Ameneh,
Thank you for your changes, and your hard work! It definitely looks much better, but I can still see two of the problems. I leave them in these comments:
- https://github.com/CodeYourFuture/Module-Tools/pull/49#discussion_r2106163155
- https://github.com/CodeYourFuture/Module-Tools/pull/49/files#r2091601751

I think that when you've addressed those, the PR will be good to go!

illicitonion:

PR #51, opened 2025-03-21
LONDON SDC | Anna Fedyna | Module-Tools | Week 4 | Implement Shell Tools in Python

annafedyna:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -11,14 +11,19 @@ parser.add_argument("paths", nargs="+", help="The file to search")
 args = parser.parse_args()
 
 for path in args.paths:
-    with open(path, "r") as f:
-        content = f.read()
-    if args.n or args.b :
-        counter = 1
-        arr = content.split('\n')[:-1]
-        for line in arr:
-            if args.b and not line:
-                print(line)
-                continue
-            print(counter, line)
-            counter += 1
+    try:
+        with open(path, "r") as f:
+            content = f.read()
+        if args.n or args.b :
+            counter = 1
+            arr = content.split('\n')[:-1]
+            for line in arr:
+                if args.b and not line:
+                    print(line)
+                    continue
+                print(counter, line)
+                counter += 1
+    except FileNotFoundError:
+        print(f"Error: File '{path}' not found.")
+    except Exception as e:
+        print(f"An error occurred: {e}")
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,0 +1,24 @@
+import argparse
+
+parser = argparse.ArgumentParser(
+                    prog='Cat Command',
+                    description='Display and concatenate files contents')
+
+parser.add_argument("-n", help="Number the output lines, starting at 1.", default = None, action="store_true")
+parser.add_argument("-b", help="Number the non-blank output lines, starting at 1.", default = None, action="store_true")
+parser.add_argument("paths", nargs="+", help="The file to search")
+
+args = parser.parse_args()
+
+for path in args.paths:
+    with open(path, "r") as f:
+        content = f.read()
+    if args.n or args.b :
+        counter = 1
+        arr = content.split('\n')[:-1]
+        for line in arr:
+            if args.b and not line:
+                print(line)
+                continue
+            print(counter, line)
+            counter += 1
--- a/implement-shell-tools/ls/ls.py
+++ b/implement-shell-tools/ls/ls.py
@@ -13,24 +13,33 @@ args = parser.parse_args()
 cwd = os.getcwd()
 
 def list_docs(dir_to_procccess):
-    documents = [document for document in os.listdir(dir_to_procccess)]
-    doc_to_output = []
-    for doc in documents:
-            if not args.a and doc.startswith("."): 
-                continue
-            else:
-                doc_to_output.append(doc)
-    doc_to_output.sort()
-    if args.one:
-        for doc in doc_to_output:
-            print(doc)
-    else:
-        print(' '.join(doc_to_output))
-
+    try:
+        documents = [document for document in os.listdir(dir_to_procccess)]
+        doc_to_output = []
+        for doc in documents:
+                if not args.a and doc.startswith("."): 
+                    continue
+                else:
+                    doc_to_output.append(doc)
+        doc_to_output.sort()
+        if args.one:
+            for doc in doc_to_output:
+                print(doc)
+        else:
+            print(' '.join(doc_to_output))
+    except FileNotFoundError:
+        print(f"Error: Directory '{dir_to_procccess}' not found.")
+    except NotADirectoryError:
+        print(f"Error: '{dir_to_procccess}' is not a directory.")
+    except Exception as e:
+        print(f"An unexpected error occurred: {e}")
 
 if args.paths:
     for directory in args.paths:
-        dir_to_procccess = os.path.join(cwd, directory)
-        list_docs(dir_to_procccess)
+        try:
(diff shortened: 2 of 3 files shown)


implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,29 @@
+import argparse
+
+parser = argparse.ArgumentParser(
+                    prog='Cat Command',
+                    description='Display and concatenate files contents')
+
+parser.add_argument("-n", help="Number the output lines, starting at 1.", default = None, action="store_true")
+parser.add_argument("-b", help="Number the non-blank output lines, starting at 1.", default = None, action="store_true")
```
blorente, line 8: What happens if neither `-n` nor `-b` are used?
For instance, what will your solution print if we invoke it like this: `python3 cat.py sample-files/1.txt`?

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,29 @@
...
+            counter = 1
+            arr = content.split('\n')[:-1]
+            for line in arr:
+                if args.b and not line:
+                    print(line)
+                    continue
+                print(counter, line)
+                counter += 1
+    except FileNotFoundError:
+        print(f"Error: File '{path}' not found.")
```
blorente, line 27: Very good catching of the error cases, this really improves the user's experience. Congrats!

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,72 @@
...
+
+args = parser.parse_args()
+cwd = os.getcwd()
+
+
+def count_lines(content):
+    return len(content.splitlines())
+    
+def count_words(content):
+    return len(re.findall("[a-zA-Z\-\.'/]+", content))
```
blorente, line 22: Whenever I run this solution, I see the following warning:
```
implement-shell-tools/wc/wc.py:22: SyntaxWarning: invalid escape sequence '\-'
  return len(re.findall("[a-zA-Z\-\.'/]+", content))
```

While the code seems to work even with the warning, could you elaborate on what could be the reason for the warning?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,72 @@
...
+                total_words += dict_output['words'] if args.words or no_options else 0
+                total_bytes += dict_output['bytes'] if args.bytes or no_options else 0
+        elif os.path.isdir(file_path_to_procccess):
+            print(f"{path}: Is a directory")
+    except FileNotFoundError:
+        print(f"Error: File '{file_path_to_procccess}' not found.")
+    except Exception as e:
+            print(f"An unexpected error occurred: {e}")
+
+if len(args.paths) > 1:
```
blorente, line 64: When I pass more than one line and any flag, I get the following error: `An unexpected error occurred: name 'no_options' is not defined`.

For instance, with this invocation:
```
➜ python3  wc.py -l sample-files/*
1 sample-files/1.txt
An unexpected error occurred: name 'no_options' is not defined
1 sample-files/2.txt
An unexpected error occurred: name 'no_options' is not defined
5 sample-files/3.txt
An unexpected error occurred: name 'no_options' is not defined
Traceback (most recent call last):
  File "/Users/blorente/code/github.com/CodeYourFuture/Module-Tools/implement-shell-tools/wc/wc.py", line 65, in <module>
    if no_options:
       ^^^^^^^^^^
NameError: name 'no_options' is not defined
```

What could be the cause of this?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,72 @@
...
+        output['words'] = count_words(content)
+    if args.bytes:
+        output['bytes'] = count_bytes(content)
+    print(' '.join([ str(val) for val in output.values()]), path)
+    return output
+
+total_lines, total_words, total_bytes = 0, 0, 0
+for path in args.paths:
+    try:
+        file_path_to_procccess = os.path.join(cwd, path)
```
blorente, line 49:
```suggestion
        file_path_to_proccess = os.path.join(cwd, path)
```
Typo: There was an extra c :D

This is not relevant for the code working, but in a real code review it's likely to get pointed out, so I do it here as well. If you decide to change it, note that you'll have to change every place where that variable is used.

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,72 @@
...
+            with open(file_path_to_procccess, 'r') as f:
+                file_content = f.read()
+                dict_output = process_file_content(file_content, path)
+                total_lines += dict_output['lines'] if args.lines or no_options else 0
+                total_words += dict_output['words'] if args.words or no_options else 0
+                total_bytes += dict_output['bytes'] if args.bytes or no_options else 0
+        elif os.path.isdir(file_path_to_procccess):
+            print(f"{path}: Is a directory")
+    except FileNotFoundError:
+        print(f"Error: File '{file_path_to_procccess}' not found.")
```
blorente, line 60: What happens if I pass three files, but the second one is missing? How will that affect the output?
For a stretch goal: How would you make it so that all errors are displayed at the end, instead of intermingled throughout the output?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,45 @@
+import argparse
+import os 
+
+parser = argparse.ArgumentParser(
+                    prog='ls command',
+                    description='list directory contents')
+
+parser.add_argument("-1", "--one", help="Force output to be one entry per line.", default = None, action="store_true")
+parser.add_argument("-a", help="Include directory entries whose names begin with a dot (.).", default = None, action="store_true")
```
blorente, line 9: When I run the `ls` utility with `-a`, the current directory (`.`) and parent directory (`..`) are printed:

```shell
$ ls -a sample-files
.               ..              .hidden.txt     1.txt           2.txt           3.txt           dir
```

How could you change your solution to make sure they are printed?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,45 @@
...
+
+if args.paths:
+    for directory in args.paths:
+        try:
+            dir_to_procccess = os.path.join(cwd, directory)
+            list_docs(dir_to_procccess)
+        except Exception as e:
+            print(f"An unexpected error occurred: {e}")
+else:
+    list_docs(cwd)
```
blorente, line 45: Good catch on the behaviour of `ls` without arguments! Well done.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,45 @@
...
+                if not args.a and doc.startswith("."): 
+                    continue
+                else:
+                    doc_to_output.append(doc)
+        doc_to_output.sort()
+        if args.one:
+            for doc in doc_to_output:
+                print(doc)
+        else:
+            print(' '.join(doc_to_output))
```
blorente, line 29: In ls, files are printed with a certain amount of padding, so that if we have many files, they appear to be in a table:
```
➜ /bin/ls
ls.js                   package-lock.json           README.md
node_modules            package.json                sample-files
```
This is not part of the assignment brief but, as a streacth goal, would you like to consider how to pad the outputs like that?

blorente (commented):
Thank you for your PR! I'm Borja, a volunteer, and I'll be reviewing your code.

The implementations of cat, ls, and wc are good! I have left a few edge cases that are not covered, as well as a few suggestions, but you're definitely on the right track :)

Keep up the good work!

PR #120, opened 2025-07-23
London | 25-SDC-July | Aung Ye Kyaw | sprint 1 | Number System Exercise 

sarawone:
Hello!

I wish you a good day.
I would like to thank you for the time to review my coursework.
Kindly check and provide feedback & guide.. 

Thanks ,
AYK

### DIFF
--- a/number-systems/README.md
+++ b/number-systems/README.md
@@ -5,61 +5,61 @@ Do not convert any binary numbers to decimal when solving a question unless the 
 The goal of these exercises is for you to gain an intuition for binary numbers. Using tools to solve the problems defeats the point.
 
 Convert the decimal number 14 to binary.
-Answer:
+Answer:1110
 
 Convert the binary number 101101 to decimal:
-Answer:
+Answer:45
 
 Which is larger: 1000 or 0111?
-Answer:
+Answer:1000
 
 Which is larger: 00100 or 01011?
-Answer:
+Answer:01011
 
 What is 10101 + 01010?
-Answer:
+Answer:11111
 
 What is 10001 + 10001?
-Answer:
+Answer:100010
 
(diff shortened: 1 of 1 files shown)


number-systems/README.md
```diff
@@ -5,61 +5,61 @@ Do not convert any binary numbers to decimal when solving a question unless the
...
-Answer:
+Answer:2bits
 
 How many bits would you need in order to store the numbers between 0 and 1000 inclusive?
-Answer:
+Answer:10 bits
 
 How can you test if a binary number is a power of two (e.g. 1, 2, 4, 8, 16, ...)?
-Answer:
+Answer: please explain for me this . Thank you.
```
LonMcGregor, line 38: Here, the goal is to find something out about a specific set of numbers. What if you try comparing some normal binary numbers (7, 11, 20...), and some binary numbers that are powers of two (i.e. 2*2, 2*2*2, etc). Can you see any patterns?

number-systems/README.md
```diff
@@ -5,61 +5,61 @@ Do not convert any binary numbers to decimal when solving a question unless the
...
-Answer:
+Answer: !
 
 If reading the byte 0x21 as a greyscale colour, as described in "Approaches for Representing Colors and Images", what colour would it mean?
-Answer:
+Answer:gray
 
 If reading the bytes 0xAA00FF as an RGB colour, as described in "Approaches for Representing Colors and Images", what colour would it mean?
-Answer:
+Answer: No idea . Please help me on this
```
LonMcGregor, line 62: Do you remember how RGB colours work? If you wanted to turn a colour, like "green" into a hex number like this, how would you do that? Once you figure that out, can you go the other way and turn hex back into a colour?

LonMcGregor (commented):
Good answers for most of these. I've left some tips for the ones you were stuck on

PR #126, opened 2025-07-28
London | 25-SDC-July | Andrei Filippov | Sprint 3 | Implement shell tools

Droid-An:
## Learners, PR Template

Self checklist

## Changelist

Done shell tools

## Questions

Are my implementations alright?

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,7 +1,5 @@
 import { promises as fs } from "node:fs";
-// import path from "node:path";
 import { program } from "commander";
-import { clearLine } from "node:readline";
 
 program
   .name("cat")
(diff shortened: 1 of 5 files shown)


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,60 @@
...
+
+const opts = program.opts();
+
+const countLines = (data) => {
+  const lines = data.split("\n");
+  if (lines[lines.length - 1] === "") {
+    lines.pop();
+  }
+
+  let lineNum = 1;
```
LonMcGregor, line 22: If I pass multiple files in, the line numbering resets to 1 for each file. Can you make your implementation work more like the original cat which uses a continuous numbering?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+};
+
+const flag_a = (files) => {
+  try {
+    files.unshift("..");
+    files.unshift(".");
+    return files;
+  } catch (err) {
+    console.error("Error reading directory:", err);
```
LonMcGregor, line 65: You have an error condition here if there was a problem reading the directory. Under what conditions can that error happen here?

(I have the same comment for the other flag)

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,76 @@
...
+  if (argv.length > 1) {
+    if (total.length == 0) {
+      total.push(...output);
+    } else {
+      for (let index = 0; index < output.length; index++) {
+        total[index] += output[index];
+      }
+    }
+  }
+  console.log([...output, path].join("     "));
```
LonMcGregor, line 62: Can you think of a way to make the files align neatly, as if in a table, when multiple files of different word counts are present?

LonMcGregor (commented):
Great work on these tasks, I have left some comments for improvements that could be made

implement-shell-tools/wc/wc.js
Droid-An: yes, I can use padStart and length of the biggest value in a output as a length argument

LonMcGregor:
Great. You're done with this task now!

Droid-An:
Awesome, thank you for the review

illicitonion:

PR #135, opened 2025-07-31
London | 25-SDC-July | Franklin Kamela | Sprint 3 | Feature/Implement Shell Tools

Fradoka:
## Learners, PR Template

Self checklist

-  I have committed my files one by one, on purpose, and for a reason
-  I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
-  I have tested my changes
-  My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)
-  My changes meet the [requirements](./README.md) of this task

### DIFF
--- a/implement-shell-tools/cat/my_cat.js
+++ b/implement-shell-tools/cat/my_cat.js
@@ -1,0 +1,41 @@
+import { promises as fs } from 'node:fs';
+import { argv } from 'node:process';
+
+const args = argv.slice(2); // skip node and script name
+
+let showLineNumb = false;
+let showNoBlankNumb = false
+let file = '';
+
+
+args.forEach(arg => {
+  if (arg === '-n') {
+    showLineNumb = true;
+  } else if (arg === '-b') {
+    showLineNumb = false;
+    showNoBlankNumb = true;
+  }else {  
+    file = arg; 
+  }
+});
+
+try {
+  let content = await fs.readFile(file, 'utf-8');
+  const lines = content.split('\n');
+
+  if (showLineNumb) {
+    // Number all lines
+    content = lines.map((line, i) => `${i + 1}\t${line}`).join('\n');
+  }else  if (showNoBlankNumb) {
+    // Number only non-blank lines
+    let counter = 1;
+    content = lines.map(line => {
+      if (line.trim() === '') return '';
+      return `${counter++}\t${line}`;
+    }).join('\n');
+  }
+
+  console.log (content)
+} catch (err) {
+  console.error(`Cannot access '${file}': ${err.message}`);
+}
--- a/implement-shell-tools/cat/my_cat.js
+++ b/implement-shell-tools/cat/my_cat.js
@@ -5,37 +5,39 @@ const args = argv.slice(2); // skip node and script name
 
 let showLineNumb = false;
 let showNoBlankNumb = false
-let file = '';
+const files = [];
 
 
 args.forEach(arg => {
   if (arg === '-n') {
     showLineNumb = true;
   } else if (arg === '-b') {
-    showLineNumb = false;
     showNoBlankNumb = true;
   }else {  
-    file = arg; 
+    files.push(arg); 
   }
 });
 
-try {
-  let content = await fs.readFile(file, 'utf-8');
-  const lines = content.split('\n');
+if (files.length === 0) {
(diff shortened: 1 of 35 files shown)


implement-shell-tools/cat/my_cat.js
```diff
@@ -0,0 +1,41 @@
...
+
+
+args.forEach(arg => {
+  if (arg === '-n') {
+    showLineNumb = true;
+  } else if (arg === '-b') {
+    showLineNumb = false;
+    showNoBlankNumb = true;
+  }else {  
+    file = arg;
```
LonMcGregor, line 18: What happens if I want to pass multiple files in here? For example, using the * wildcard which expands to multiple files in the shell. 

Your solution in `wc` seems a better approach

implement-shell-tools/cat/my_cat.js
```diff
@@ -0,0 +1,41 @@
...
+  }else  if (showNoBlankNumb) {
+    // Number only non-blank lines
+    let counter = 1;
+    content = lines.map(line => {
+      if (line.trim() === '') return '';
+      return `${counter++}\t${line}`;
+    }).join('\n');
+  }
+
+  console.log (content)
```
LonMcGregor, line 38: Be careful of formatting your code the same way. Can you see why this line is inconsistent with the rest of the file?

implement-shell-tools/ls/my_ls.js
```diff
@@ -0,0 +1,29 @@
...
+
+let showAll = false;
+let dir = '.'; //set dir to current directory
+
+// Parse args to check weather or not to show hidden files
+//and ignore -1 because our output is already 1 per line
+args.forEach(arg => {
+  if (arg === '-a') {
+    showAll = true;
+  } else if (arg !== '-1') {
```
LonMcGregor, line 14: This approach might cause problems in two cases:
- What if the user does not want each output per line and wants it all on one line (like the original `ls`) app?
- Assuming anything else is a directory might make it more difficult to add new options in the future.

implement-shell-tools/wc/my_wc.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const validCounts = counts.filter(c => c !== null);
+
+  if (validCounts.length > 1) {
+    const totalLines = validCounts.reduce((sum, c) => sum + c.lines, 0);
+    const totalWords = validCounts.reduce((sum, c) => sum + c.words, 0);
+    const totalBytes = validCounts.reduce((sum, c) => sum + c.bytes, 0);
+
+    let totalOutput = '';
+    if (showLines) totalOutput += `${totalLines} `;
```
LonMcGregor, line 79: There is some duplication here when you are printing totals and when you print single file output. Can you think of how to reduce this?

implement-shell-tools/wc/my_wc.js
```diff
@@ -0,0 +1,89 @@
...
+
+    return { lines, words, bytes };
+  } catch (err) {
+    console.error(`Cannot access '${file}': ${err.message}`);
+    return null;
+  }
+}
+
+async function main() {
+  const counts = await Promise.all(files.map(countFile));
```
LonMcGregor, line 69: This is a very minor point, but when doing `map` we expect it to return a new array that is different from the input array. Instead of `.map` can you think of another array function that would make more sense to use here?

LonMcGregor (commented):
Hi,

Remember to update the template when you open a new PR on github - you have left the default text there, so I don't know what the goal of this PR is.

You have made a good start on these tasks. I've left some comments for how you could improve them further.

Fradoka:
hi @LonMcGregor, i have now amended this cat, ls, and wc to address review.

LonMcGregor (commented):
Good work on updating these. The only thing left to work in is making sure that your `cat` can handle numbering multiple files

Fradoka:
hi @LonMcGregor, amended my_cat.js to address review.

Fradoka:
hi @LonMcGregor, i don't actually need 2 different counter variables as `-b` will overide `-n` so there won't be a time when i would need both set of numbering. i used 2 counters before to differentiate but a unique counter still does the job. i have now amended the code to address that.

illicitonion:

PR #138, opened 2025-08-01
London | 25-SDC-July | Fatma Arslantas | Sprint 4 | Implement Shell Tools Python

AFatmaa:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -29,10 +29,10 @@ parser = argparse.ArgumentParser(
     description="Display contents of files with optional line numbering."
 )
 
-group = parser.add_mutually_exclusive_group()
+numbering_group = parser.add_mutually_exclusive_group()
 
-group.add_argument("-n", action="store_true", help="Number all lines")
-group.add_argument("-b", action="store_true", help="Number non-blank lines")
+numbering_group.add_argument("-n", action="store_true", help="Number all lines")
+numbering_group.add_argument("-b", action="store_true", help="Number non-blank lines")
 parser.add_argument("files", nargs="+", help="Files to display")
 
 args = parser.parse_args()
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -21,19 +21,16 @@ number_non_blank = args.b
 for file_path in args.files:
     with open(file_path, "r", encoding="utf-8") as file:
         lines = file.readlines()
-        line_num = 1  # Start from 1
 
-        for line in lines:
+        for line_num, line in enumerate(lines, start=1):
             # -n: number all lines
             if number_all_lines:
                 print(f"{line_num:6}\t{line}", end="") # 'line_num:6' means align right
-                line_num += 1
 
             # -b: number only non-blank lines
             elif number_non_blank:
                 if line.strip() != "":
                     print(f"{line_num:6}\t{line}", end="")
-                    line_num += 1
                 else:
                     print(line, end="")
 
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,5 +1,23 @@
 import argparse
 
+def cat_file(file_path, number_all_lines, number_non_blank):
+  with open(file_path, "r", encoding="utf-8") as file:
+        lines = file.readlines()
+
+        for line_num, line in enumerate(lines, start=1):
+            prefix = ""
+
+            # -n: number all lines
+            if number_all_lines:
+                prefix = f"{line_num:6}\t" # 'line_num:6' means align right
+
+            # -b: number only non-blank lines
+            elif number_non_blank and line.strip() != "":
+                    prefix = f"{line_num:6}\t"
+
+            print(f"{prefix}{line}", end="")
+            
+
 parser = argparse.ArgumentParser(
     prog="cat.py",
     description="Display contents of files with optional line numbering."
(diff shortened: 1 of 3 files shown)


implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,40 @@
...
+
+parser.add_argument("-n", action="store_true", help="Number all lines")
+parser.add_argument("-b", action="store_true", help="Number non-blank lines")
+parser.add_argument("files", nargs="+", help="Files to display")
+
+args = parser.parse_args()
+
+# Check if user used -n or -b
+number_all_lines = args.n
+number_non_blank = args.b
```
DaryaShirokova, line 16: Non-blocking: it looks like if both -n and -b are specified, -b doesn't have any affect. It could be added as mutually exclusive flags: https://docs.python.org/3/library/argparse.html#mutual-exclusion

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,40 @@
...
+
+# Loop through all file paths
+for file_path in args.files:
+    with open(file_path, "r", encoding="utf-8") as file:
+        lines = file.readlines()
+        line_num = 1  # Start from 1
+
+        for line in lines:
```
DaryaShirokova, line 24: You could use [enumerate](https://docs.python.org/3/library/functions.html#enumerate) here to not count line numbers manually.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,40 @@
...
+            # -n: number all lines
+            if number_all_lines:
```
DaryaShirokova, line 26: The structure of if/else could be slightly simpler if you store the prefix you want to print in a variable, something like:
```
prefix = ""
if number_all_lines:
    prefix = ...
elif number_non_blank and line.strip() != "":
    prefix = ...
print(f"{prefix}{line}", end="")
```

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,40 @@
...
+
+args = parser.parse_args()
+
+# Check if user used -n or -b
+number_all_lines = args.n
+number_non_blank = args.b
+
+# Loop through all file paths
+for file_path in args.files:
+    with open(file_path, "r", encoding="utf-8") as file:
```
DaryaShirokova, line 20: This could optionally be moved to a helper function, such as `cat_file(file_path)`.

DaryaShirokova, line 20: Here and in another scripts, you could check that the file actually exists and print a user friendly message if it doesn't.

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,66 @@
...
+        output.append(str(word_count))
+    if args.c:
+        output.append(str(char_count))
+
+    # If no flag is given, show all
+    if not (args.l or args.w or args.c):
+        output = [str(line_count), str(word_count), str(char_count)]
+
+    output.append(file_path)
+    print(" ".join(output))
```
DaryaShirokova, line 49: This can be moved to a helper function, something like `format_wc_output` and reused for the total output as well (if parametrized).

DaryaShirokova (approved):
Looks good! I left a few comments for some potential improvements, but they are non-blocking.

implement-shell-tools/cat/cat.py
AFatmaa: Hi @DaryaShirokova,

Thank you for the feedback! I've implemented the mutually exclusive flags for -n and -b as suggested. Now it correctly prevents using both at the same time 🙏

implement-shell-tools/cat/cat.py
AFatmaa: I wasn't aware of `enumerate()` for this use case, but it definitely makes the line counting cleaner. I've updated the code accordingly. Thanks a lot!

implement-shell-tools/cat/cat.py
AFatmaa: I've refactored the print statements according to your suggestions and updated the code. It reduces repetition. Thank you!

implement-shell-tools/wc/wc.py
AFatmaa: I've refactored the output formatting into a `format_output` helper function to reduce duplication. I also added error handling to check for non-existent files and print user-friendly messages. Thanks for guiding me!

AFatmaa:
Hi @DaryaShirokova,

I've implemented all the suggestions you provided. Your feedback was helpful, and I've learned a lot. Thank you so much for your guidance and your time! 🌸

Please let me know if there's anything else I need to adjust 🙏

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,46 @@
...
+    except Exception as e:
+        print(f"cat: {file_path}: Error reading file: {e}", file=sys.stderr)
+
+
+parser = argparse.ArgumentParser(
+    prog="cat.py",
+    description="Display contents of files with optional line numbering."
+)
+
+group = parser.add_mutually_exclusive_group()
```
DaryaShirokova, line 32: Nit, you don't need to fix this, just for future reference. I would give slightly more explicit name like numbering_group or something similar just for better readability.

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,69 @@
...
+
+parser.add_argument("-l", action="store_true", help="Count lines")
+parser.add_argument("-w", action="store_true", help="Count words")
+parser.add_argument("-c", action="store_true", help="Count characters")
+
+parser.add_argument("files", nargs="+", help="Text files to read")
+
+args = parser.parse_args()
+
+def format_output(line_count, word_count, char_count, args):
```
DaryaShirokova, line 17: Also nit, optional - it works in the context of this task. Overall I wouldn't pass args to a method definition as it is not really clear how exactly it is supposed to look without checking into the implementation details. Something with clear name like count_lines, count_words etc might work better.

DaryaShirokova (approved):
LGTM!

implement-shell-tools/cat/cat.py
AFatmaa: You're right, readability is definitely important! I've updated the variable name. Thanks for pointing that out and helping me improve!

implement-shell-tools/wc/wc.py
AFatmaa: I see what you mean. Passing specific flags instead of the whole `args` object definitely makes the function's purpose clearer and more understandable. I've updated the function. Thanks for the guidance!

illicitonion:

PR #144, opened 2025-08-03
West Midlands | SDC-July | Gabriel Deng | Sprint 4 | Implement shell tools in python

gai93003:
Having people write `s/s/S/` is a little more confusing than `s/i/I/`
for their first exercise.

## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


LonMcGregor:
Hi, I can't review your code yet. If you look at the "files changed" tab on this PR you will see the problem as there are hundreds of files that don't need to be there.

It looks like you're using a virtual environment, which is a good idea if you are using multiple libraries, but needs to be handled carefully within a VCS. Could you look up how to properly handle python virtual environments within git so this problem doesn't happen?

gai93003:
> That's better - only the files relevant are committed. In future be careful of committing the compiled python binaries, they don't need to be there.

Thanks, I was not yet familiar with the running python. I will look out for these things in the future

illicitonion:

PR #148, opened 2025-08-05
ZA | 25-SDC-July | Luke Manyamazi | Sprint 4 | Python Implement Shell Tools Exercises

### LINKED ISSUE #177: 01 Implement shell tools (cat, ls, wc) in Python
### Link to the coursework

https://github.com/CodeYourFuture/Module-Tools/tree/main/implement-shell-tools

You should implement these tools in Python.

### Why are we doing this?

Practicing designing and writing code, getting used to writing small tools, and understanding argument parsing.

### Maximum time in hours

9

### How to submit

Submit a PR from your fork of the Module-Tools repo with your implementations.

Luke-Manyamazi:
## Learners, PR Template

Self checklist

## Changelist

This PR implements basic versions of the following Unix commands in Python:

- cat: Concatenates and displays file contents, with support for -n (number all lines) and -b (number non-empty lines).
- ls: Lists directory contents, supporting options for showing hidden files and listing in one column.
- wc: Counts lines, words, and bytes in a file, with flags to show specific counts.

These scripts were designed to mimic standard command-line tools and handle typical edge cases like missing files and glob patterns.

## Questions

1. Are there any improvements you would suggest for handling edge cases (e.g., invalid file paths, empty files)?
2. Is the use of argparse and glob in each script appropriate and idiomatic?
3. Would you recommend any refactoring for readability or modularity?
4. Are there any Python best practices I may have missed?

### DIFF
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,17 +1,23 @@
 import glob
 import argparse
+from enum import Enum
 
-def cat(filepath, n=False, b=False, line_counter=None):
+class Numbering(Enum):
+    NONE = 0
+    ALL = 1
+    NONEMPTY = 2
+
+def cat(filepath, numbering=Numbering.NONE, line_counter=None):
     try:
         with open(filepath) as f:
             for line in f:
-                if b:
+                if numbering == Numbering.NONEMPTY:
                     if line.strip():
                         print(f"{line_counter[0]:6}\t{line}", end='')
                         line_counter[0] += 1
                     else:
                         print(line, end='')
-                elif n:
+                elif numbering == Numbering.ALL:
                     print(f"{line_counter[0]:6}\t{line}", end='')
                     line_counter[0] += 1
                 else:
@@ -20,7 +26,7 @@ def cat(filepath, n=False, b=False, line_counter=None):
         print(f"cat: {filepath}: No such file or directory")
 
 def main():
-    parser = argparse.ArgumentParser(description = "Concatenate files and print on the standard output.")
+    parser = argparse.ArgumentParser(description="Concatenate files and print on the standard output.")
     parser.add_argument('-n', action='store_true', help='number all output lines')
     parser.add_argument('-b', action='store_true', help='number non-empty output lines')
     parser.add_argument('files', nargs='+', help='files to concatenate')
@@ -31,8 +37,19 @@ def main():
         files.extend(glob.glob(pattern) or [pattern])
 
     line_counter = [1]
+
+    # Determine numbering mode (mutually exclusive)
+    if args.n and args.b:
+        parser.error("options -n and -b are mutually exclusive")
+    elif args.n:
+        numbering = Numbering.ALL
+    elif args.b:
+        numbering = Numbering.NONEMPTY
+    else:
+        numbering = Numbering.NONE
+
     for file in sorted(files):
-        cat(file, args.n, args.b, line_counter)
+        cat(file, numbering=numbering, line_counter=line_counter)
 
 if __name__ == "__main__":
     main()
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,4 +1,3 @@
-import glob
 import argparse
 from enum import Enum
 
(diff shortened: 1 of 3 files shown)


illicitonion (commented):
This is generally looking good, but I left a few things to think about :)

illicitonion (commented):
This is generally looking good, but there are a bunch of small details and consistencies to look at :)

Luke-Manyamazi:
Thanks for the review. I've addressed the remaining comments by:

* Sending error messages to stderr in cat, ls and wc.
* Updating ls to handle file paths and include "." and ".." when using -a.
* Re-testing cat and ls behaviour against the corresponding Unix commands.
* Simplifying variable initialisation and duplicated conditions in wc.
* Returning full counts from wc() and letting the caller determine what to display.
* Making count handling more consistent throughout wc.

Please let me know if there is anything else that should be adjusted.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,59 @@
...
+    try:
+        with open(filepath) as f:
+            for line in f:
+                if numbering == Numbering.NONEMPTY:
+                    if line.strip():
+                        line_number = print_numbered_line(line, line_number)
+                    else:
+                        print(line, end='')
+                elif numbering == Numbering.ALL:
+                    line_number = print_numbered_line(line, line_number)
```
illicitonion, line 25: Personally I would probably to make the `if` more general here and write this as:


```suggestion
                if numbering == Numbering.ALL or (numbering == Numbering.NONEMPTY and line.strip()):
                    line_number = print_numbered_line(line, line_number)
                else:
                    print(line, end='')
```

This is because the _reason_ we're picking the two identical behaviours is the same ("We should number this line"), so we ideally only want to write the behaviour once.

And once we've done this, because there's no repetition, I'd probably stop having `print_numbered_line` return a new number, and instead just:

```suggestion
                if numbering == Numbering.ALL or (numbering == Numbering.NONEMPTY and line.strip()):
                   print_numbered_line(line, line_number)
                   line_number += 1
                else:
                    print(line, end='')
```

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,47 @@
...
+            files = ['.', '..'] + files
+        else:
+            files = [f for f in files if not f.startswith('.')]
+
+        files.sort()
+
+        if one_column:
+            print(*files, sep='\n')
+        else:
+            print(*files)
```
illicitonion, line 24: This is fine, but an alternative you may want to consider is:

```suggestion
        separator = "\t"
        if one_column:
            separator = "\n"

        print(*files, sep=separator)
```

This may be more or less clear, I'm not going to argue strongly one way or the other, but is in keeping with that general "highlight what's the same and what's different" idea.

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,87 @@
...
+
+    except FileNotFoundError:
+        print(
+            f"wc: {path}: No such file or directory",
+            file=sys.stderr
+        )
+        return (0, 0, 0)
+
+    except IsADirectoryError:
+        print(
```
illicitonion, line 45: If this happens, what will the exit code of the process be? What _should_ it be?

(Same question for `FileNotFoundError`)

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,87 @@
...
+        if args.l or show_all:
+            parts.append(str(total_lines))
+
+        if args.w or show_all:
+            parts.append(str(total_words))
+
+        if args.c or show_all:
+            parts.append(str(total_bytes))
+
+        print(' '.join(parts), 'total')
```
illicitonion, line 84: This is repetitive with the formatting code, but as it's currently structured it's quite fiddly to deduplicate. Here's a set of steps that may help:
1. Move the "if none of the flags are set, set them all to true" code into just after `args = parser.parse_args`. Actually update `args.l` and `args.w` and `args.c` as a result. Then you don't need the `or show_all` any more.
2. Extract a function `def print_stats(args, lines, words, bytes, filename)` which you can call from both places.

illicitonion (commented):
Sorry for the _huge_ delay in getting back to you on this. Hopefully one last set of comments, which are hopefully useful to think about!

Luke-Manyamazi:
Hi @illicitonion, I’ve addressed the review comments in the three Python implementations:

- `cat.py`: improved line-numbering logic, `-n`/`-b` exclusivity, stderr errors, directory handling, and non-zero exit codes.
- `ls.py`: removed unnecessary function defaults, handled files and invalid paths, included `.` and `..` with `-a`, used stderr, and refactored output formatting.
- `wc.py`: simplified count selection, extracted shared formatting, added multi-file totals, handled errors through stderr, and returned non-zero exit codes for failures.

I tested the required single-file, multi-file, flag, invalid-path, and syntax-check cases. Please could you review the latest changes and let me know whether anything else needs adjusting?

illicitonion (commented):
Looks good, well done!

PR #157, opened 2025-08-08
London | 25-SDC-July | Eyuel Abraham | Sprint 4 | Implementing cowsay

eyuell21:
## Learners, PR Template

Self checklist

## Changelist

Python implementation of Cowsay.

## Questions


implement-cowsay/cow.py
```diff
@@ -0,0 +1,50 @@
...
+        default='cow',
+        help=f'The animal to be saying things.'
+    )
+
+    args = parser.parse_args()
+    message = ' '.join(args.message)
+
+    # Dynamically call the animal function (like cowsay.turtle("hello"))
+    try:
+        animal_func = getattr(cowsay, args.animal)
```
DaryaShirokova, line 42: You can check out an alternative `get_output_string` method here to avoid `getattr` (not the best choice to use if there is alternative - it make code more readable).

implement-cowsay/cow.py
```diff
@@ -0,0 +1,50 @@
...
+        output = animal_func(message)
+        print(output)
```
DaryaShirokova, line 44: You program always prints none in the end, please debug why this is and fix (should be a simple fix :) )

implement-cowsay/cow.py
eyuell21: Noted, Thank you 🙏

implement-cowsay/cow.py
```diff
@@ -41,7 +41,8 @@ def main():
     try:
         animal_func = getattr(cowsay, args.animal)
         output = animal_func(message)
-        print(output)
+        if output is not None:
```
DaryaShirokova: So can you explain in which cases the output is `None` here and when it is not `None`?

eyuell21: Output is None only if the animal function doesn't return anything. In the official cowsay module, it always returns a string.

The output is not None in normal cases because standard cowsay animal functions return a string of ASCII art.

illicitonion: This questoin is still open @eyuell21

eyuell21: When animal_func() returns None, print(None) outputs the literal string "None". The if check prevents printing when there's no actual output.

illicitonion:

PR #162, opened 2025-08-13
London | 25-SDC-July | Priscilla Emebo | Sprint 4 | Python Implementation

Priscilla-EM:
## Learners, PR Template

Self checklist

## Changelist

Implementation of custom wc command in Python to mirror the behavior of the real Unix wc command.

This PR implements a Python version of the Unix wc (word count) command.

The script supports counting:
Lines with the -l flag
Words with the -w flag
Bytes with the -c flag

It also supports multiple files and displays a total line when more than one file is provided.

I verified the script works correctly for the following commands as required in the README:

wc sample-files/*
wc -l sample-files/3.txt
wc -w sample-files/3.txt
wc -c sample-files/3.txt
wc -l sample-files/*

### DIFF
--- a/implement-shell-tools/wc/mywc.py
+++ b/implement-shell-tools/wc/mywc.py
@@ -1,0 +1,72 @@
+import argparse
+import glob
+import os
+
+def count_file(file_path, count_lines=False, count_words=False, count_bytes=False):
+    lines = words = bytes_count = 0
+    with open(file_path, "r", encoding="utf-8", errors="ignore") as f:
+        for line in f:
+            lines += 1
+            words += len(line.split())
+    bytes_count = os.path.getsize(file_path)
+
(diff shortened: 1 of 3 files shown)


implement-shell-tools/wc/mywc.py
LonMcGregor, line 1: There is some duplication of code in this file. Can you spot where it is and find a way to reduce this?

LonMcGregor (commented):
Remember to update the default PR text when submitting on github.

Good work on these tasks, I only have comments for one of your files

LonMcGregor (commented):
This is an improvement, good work. always be on the lookout for ways you can abstract and pullout code into separate functions like this.

illicitonion:

PR #231, opened 2025-11-17
London | 25-SDC-Nov | Jesus del Moral | Sprint 2 | JQ

delmorallopez:
SDC Sprint 3 | jq exercises

Understanding and manipulating JSON. The best way to get to grips with jq is by using it.

jq section exercices under Sprint2

### DIFF
--- a/jq/script-08.sh
+++ b/jq/script-08.sh
@@ -7,4 +7,4 @@ set -euo pipefail
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 3" with no quotes.
 
-jq -r '.[] | .name + ", " + ((.scores | length) | tostring)' scores.json
+jq -r '.[] | .name + ", " + (.scores | length | tostring)' scores.json
--- a/jq/script-08.sh
+++ b/jq/script-08.sh
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the number of times they've played the game.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 3" with no quotes.
+
+jq -r '.[] | .name + ", " + ((.scores | length) | tostring)' scores.json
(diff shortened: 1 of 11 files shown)


LonMcGregor (commented):
Did you mean to make changes to all these files? We try to keep pull requests specific to just the files in the feature we are changing - for this, everything within the jq sprint task. I can't continue reviewing until the PR is made more specific.

jq/script-08.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the number of times they've played the game.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 3" with no quotes.
+
+jq -r '.[] | .name + ", " + ((.scores | length) | tostring)' scores.json
```
LonMcGregor: What does wrapping the middle (.scores | length) in brackets here do?

LonMcGregor (commented):
Good work, I just have one comment about some of your JQ commands

jq/script-08.sh
delmorallopez: .scores | length,  Take the array in .scores and count its elements

( ... )  Treat that whole pipeline as one single value

| tostring,  Convert that value (a number) into a string

.name + ", " + <value>,  Concatenate everything

So parentheses force jq to evaluate the inner expression first, jq requires grouping when mixing pipelines with string concatenation.

delmorallopez: running without brackets 
jq -r '.[] | .name + ", " + .scores | length' scores.json

got an error 
jq: error (at scores.json:1): string ("Ahmed, ") and array ([1,10,4]) cannot be added

That happens because | (pipe) runs after + (string concat)
jq interprets it as: (.name + ", " + .scores) | length

Yes, removing the parentheses changes the output drastically

delmorallopez: I can see what you mean
I have modified the script to (.scores | length | tostring)
Working fine

LonMcGregor (commented):
Looks good now!

illicitonion:

PR #239, opened 2025-11-20
London | 25-SDC-Nov | Zohreh Kazemianpour | Sprint 3 | Implement Shell Tools-NodeJS

zohrehKazemianpour:
Self checklist

This PR contains implementing shell tools (cat, ls, wc) in JavaScript with NodeJS using the commander library.


implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,43 @@
+import { program } from "commander";
+import { promises as fs } from "node:fs";
+
+//config the program
+program
+  .name("ls")
+  .description("list directory contents")
+  .option("-1, --one", "Force output to be one entry per line")
```
LonMcGregor: You include the 1 option here, do you ever use it?

LonMcGregor (commented):
Good work on these tasks, sorry for the delay in getting a review.

Good use of comments to explain throughout.

There are some places you could improve implementations further

implement-shell-tools/ls/ls.js
zohrehKazemianpour: Thanks for pointing that out. I realized the flag wasn't doing anything because my output was always vertical. I’ve refactored it to be horizontal by default, so the -1 flag now correctly 'forces' the single-column output.

illicitonion:

PR #240, opened 2025-11-20
London | SDC-Nov-2025| Ikenna Agulobi | Sprint 1 | Individual Shell Tools

ike-agu:
## Learners, PR Template

Self checklist

## Changelist
- ls:  Add ls command flags to list files as required in exercise
- cat: Add cat commands to output file contents as required in exercise
- wc: use wc commands to count words, lines and character in files
- grep: Add grep command to count Doctor lines across dialogue files
- sed:  use sed commands to replace,  modify and insert texts

### DIFF
--- a/individual-shell-tools/grep/script-03.sh
+++ b/individual-shell-tools/grep/script-03.sh
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the number of lines in dialogue.txt that contain the word Doctor (regardless of case).
 # The output should be exactly the number 9.
+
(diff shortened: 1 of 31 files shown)


individual-shell-tools/grep/script-03.sh
LonMcGregor: Correct, but remember you can combine some single-character options into one, rather than listing them separately

LonMcGregor (commented):
This looks good to me, see my comment in one of the files for some extra advice

illicitonion:

PR #241, opened 2025-11-20
Sheffield | 25-SDC-Nov | Sheida Shabankari | Sprint 3 |Implement shell tools

sheida-shab:
Self checklist


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,46 @@
...
+        flag=process.argv[2];
+        fileNamePattern = process.argv[3];
+     } else {
+         fileNamePattern = process.argv[2];
+       }
+    
+       const filesNameArray = await glob(fileNamePattern);
+       filesNameArray.sort();
+       let lineNumber = 1;
+       //console.log(filesNameArray);
```
LonMcGregor: Tip: Avoid committing commented old code

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,46 @@
+import { promises as fs } from "node:fs";
+import { glob } from "glob";
```
LonMcGregor: When I run this code I get an error `Cannot find package 'glob'`. Do you see this also?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,26 @@
+import { promises as fs } from "node:fs";
+async function listDir() {
+   let finalFileList;
+   const flag = process.argv[2];
```
LonMcGregor: The readme also asks you to try implementing the `-1` option, can you try this?

LonMcGregor (commented):
Good work on these tasks, apologies for the lateness in review

You have taken a good approach at manually parsing the arguments. You might consider investigating if there are any libraries or APIs you could use to help you with this. In a more complex shell tool, it might be difficult to manually check arguments this way. It would also make it easier to combine arguments, e.g.allowing `wc -w -l` instead of only one at a time

implement-shell-tools/cat/cat.js
sheida-shab: Thanks for your comment but I don't see any error !

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,46 @@
...
+            //console.log(linesOfText);
+            for(const line of linesOfText){
+                if (line.trim()==="" && flag==='-b'){
+                 console.log(line);
+                }else {
+                console.log(`${lineNumber} ${line}`);
+                lineNumber ++ ;
+                }
+            }
+            }
```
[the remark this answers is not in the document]
sheida-shab: I’m using VS Code with Prettier enabled, but I’m not sure why the file wasn’t properly formatted when saved. I manually reformatted it afterwards.

implement-shell-tools/ls/ls.js
sheida-shab: When I initially implemented the code, the -1 flag behavior was applied automatically by default, so I didn’t explicitly implement it. However, based on your feedback, I have now updated the implementation to handle the -1 flag explicitly.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,74 @@
...
+    const flags = commandLineArray.filter((item) => item.startsWith("-"));
+    
+    if(flags.length !==0 )
+    {
+        commandLineArray.shift();
+    }
+
+    for(const file of commandLineArray){
+        const fileContent=await fs.readFile(file , 'utf-8');
+        const fileBuffer=await fs.readFile(file);
```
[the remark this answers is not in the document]
sheida-shab: You're right — the file is being read twice unnecessarily, which affects performance. This could be improved by reading the file once and reusing the buffer. I updated the code.

sheida-shab:
Thanks for your feedback.I think I fixed the issue.
Could you please review it again ?

illicitonion:

PR #248, opened 2025-11-24
Sheffield | 25-SDC-Nov | Hassan Osman | Sprint 3 | Implement Shell Tools

HassanOHOsman:
Self checklist

This PR provides implementations of shell tools: cat, ls, and wc, using JavaScript with Node.js.

## Questions
Is the point of implementing those shell command as a way to appreciate the connivance they provide? 
Are there any special cases where implementing shell commands might be needed? 

Update: Having just stated Sprint 4, I now know that Shell command tools may not provide a full resolution to our problem. So to me,  this is an example of when we could resort to building our own custom tools.


MattGoodwin0 (changes requested):
You've got a good understanding of wc, ls and cat commands.

Take a look at the layout of your code files to ensure a consistent format - using a standard format will make it easier to read and review your code.

Could you add some wording around the advantage of having separate package.json files?

HassanOHOsman:
> You've got a good understanding of wc, ls and cat commands.
> 
> Take a look at the layout of your code files to ensure a consistent format - using a standard format will make it easier to read and review your code.
> 
> Could you add some wording around the advantage of having separate package.json files?

Regarding "Could you add some wording around the advantage of having separate package.json files?", was this a question in a backlog? I could not locate it if that was the case.

MattGoodwin0:
awesome!

illicitonion:

PR #250, opened 2025-11-25
London | SDC-Nov-25| Ikenna Agulobi | sprint 2 | Shell pipelines 

ike-agu:
## Learners, PR Template

Self checklist

## Changelist

This Pr contains sprint 2 shell pipeline exercise.
- ## Ls-grep exercise: 
- Listing all files containing  at least one capital letters
- Listing files starting with capital letters 
- List all names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
- Count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.

## sort-uniq-head-tail
- Sort all the scores in the files 
- Use Sort and grep to print out files 
-  Add script to count Entry and Exit events from the timestamped file.

## tr 
- Use tr to replace exclamation marks with full stops in text.txt
- Use tr to swap Y and Z (upper and lower case) in text.txt

### DIFF
--- a/shell-pipelines/ls-grep/script-03.sh
+++ b/shell-pipelines/ls-grep/script-03.sh
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+
+
--- a/shell-pipelines/ls-grep/script-03.sh
+++ b/shell-pipelines/ls-grep/script-03.sh
@@ -5,4 +5,4 @@ set -euo pipefail
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
 
-
+ls sample-files | grep '^[A-Z][^A-Z]*$'
--- a/shell-pipelines/ls-grep/script-04.sh
+++ b/shell-pipelines/ls-grep/script-04.sh
@@ -4,3 +4,6 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
+
+# ls sample-files | grep "^[A-Z][a-z]*$"
+ls sample-files | grep -c "^[A-Z][a-z]*$"
--- a/shell-pipelines/ls-grep/script-04.sh
+++ b/shell-pipelines/ls-grep/script-04.sh
@@ -5,5 +5,4 @@ set -euo pipefail
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
 
-# ls sample-files | grep "^[A-Z][a-z]*$"
-ls sample-files | grep -c "^[A-Z][a-z]*$"
+ls sample-files | grep -c "^[A-Z][^A-Z]*$"
--- a/shell-pipelines/sort-uniq-head-tail/script-02.sh
+++ b/shell-pipelines/sort-uniq-head-tail/script-02.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
 # The first line of your output should be "Basia London 22 9 6" (with no quotes).
+
+sort -k2 -r scores-table.txt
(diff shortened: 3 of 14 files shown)


shell-pipelines/ls-grep/script-03.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+
```
DaryaShirokova: I think you might have accidentally deleted the line here.

shell-pipelines/sort-uniq-head-tail/script-02.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
 # The first line of your output should be "Basia London 22 9 6" (with no quotes).
+
+sort -k2 -r scores-table.txt
```
DaryaShirokova: The output seems to be:
```
Basia London 22 9 6
Ahmed London 1 10 4
Leila London 1
Piotr Glasgow 15 2 25 11 8
Mehmet Birmingham 3 12 17
Chandra Birmingham 12 6
```

Looks like they are not sorted by score?

DaryaShirokova (commented):
Looks good! Just a few comments

shell-pipelines/ls-grep/script-03.sh
ike-agu: @DaryaShirokova thanks for pointing it out.

shell-pipelines/ls-grep/script-04.sh
```diff
@@ -4,3 +4,6 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
+
+# ls sample-files | grep "^[A-Z][a-z]*$"
+ls sample-files | grep -c "^[A-Z][a-z]*$"
```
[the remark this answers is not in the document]
ike-agu: @DaryaShirokova I've implemented the fix for this. Thanks for reviewing my PR.

shell-pipelines/sort-uniq-head-tail/script-02.sh
ike-agu: I've implemented the fix.
Thank you for the review and for pointing out the error.

ike-agu:
Hi @DaryaShirokova , thank you for taking the time to review my PR and for pointing out the necessary changes. I’ve implemented all the requested updates. When you have a moment, could you please take another look?

illicitonion (commented):
LGTM - just one question :)

ike-agu:
> LGTM - just one question :)

Please what question are you referring?

ike-agu:
@illicitonion , Thanks for reviewing my PR.  
Could you let me know what question you are referring?

shell-pipelines/sort-uniq-head-tail/script-02.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
 # The first line of your output should be "Basia London 22 9 6" (with no quotes).
+
+sort -k3,3 -nr scores-table.txt
```
illicitonion, line 9: Can you talk through what the `-k3,3` means here, and how it's different from just `-k3` that you used in script 3?

ike-agu: Hi @illicitonion , In my script above, -k3,3 means that the sort key starts and ends at column 3, i.e., only the third column is used for sorting. Meanwhile, -k3 means that the sort key starts at column 3 and continues to the end of the line, so the sorting includes column 3 and all subsequent columns.

ike-agu:
@illicitonion , I've responded to your comment on my PR. When you have a moment, could you please take a look? Thank you!

shell-pipelines/sort-uniq-head-tail/script-02.sh
illicitonion: In some of the exercises you've used `-k3` and in others `-k3,3` even though the questions are phrased the same... Do you think it matters which you use in these cases?

ike-agu: @illicitonion , Now that I’ve read your question more carefully, I think it does matter. Since the exercise says to sort by the first score, using -k3,3 is more accurate because it limits the sort to just that field. Using -k3 starts at field 3 but includes the rest of the line as part of the key. So for consistency and to match the wording of the exercise, I have updated my script-03.sh exercise to use -k3,3 instead of -k3. 
Please let me know if this reasoning is correct. Thanks!

illicitonion: I think either is probably fine here :) If there's a tie for score, we don't specify how we'd like them sorted (maybe it should be by first name, or by second score, or by the order they were in the file originally), so either `-k3` or `-k3,3` meets the requirements for the exercise :)

illicitonion:

PR #259, opened 2025-11-28
Manchester | 25-SDC-Nov | Geraldine Edwards | Sprint 4 | Implement Shell Tools Python

Geraldine-Edwards:
## Learners, PR Template

Self checklist

## Changelist

This PR contains the python exercises in the cat, ls and wc directories of the implement-shell-tools directory

## Questions

Any advice for refactoring is welcome!


implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,47 @@
+import os
+import argparse
+import locale
+
+
+#set locale to the machine default setting
+locale.setlocale(locale.LC_ALL, '')
```
OracPrime: I'm not familiar with this function.  It seems odd that you're having to set something to its default.  If it's the default, isn't it already set?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,47 @@
...
+    print(f"ls: cannot access '${args.directory}': No such file or directory")
+    exit(1)
+
+# check if path is a directory
+if os.path.isdir(args.directory):
+    entries = os.listdir(args.directory)
+
+    # if -a flag set
+    if args.all:
+        entries.extend(['.', '..'])
```
OracPrime: Doing this review without running the code, so this may be rubbish.  But . and .. are current and parent directory.  That's not the same as adding hidden files is it?  I honestly can't say this is wrong, but I need persuading it's right

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,68 @@
...
+
+# when no specific flags are set
+no_flags = not (args.lines or args.words or args.bytes)
+
+totals = {"lines": 0, "words": 0, "bytes": 0}
+
+def calculate_counts(content):
+    # returns a count object
+    return {
+        "lines": content.count('\n') + (1 if content and not content.endswith('\n') else 0),
```
OracPrime: I think the code is right, but it's a little obscure.  The right of the + probably deserves an explanatory comment

OracPrime (commented):
I can't see anything obviously wrong, but there are a few bits not obviously right that I'm asking for comments on.  So this review says (probably) more about my knowledge of how these linuxy things operate than yours.  But code should be for the reader, not the writer, so I feel justified in making my comments!

implement-shell-tools/ls/ls.py
Geraldine-Edwards: Thanks for this David :)
In my code, os.listdir() returns all files and folders in the directory, including hidden files, but it doesn't include the special directory entries (. and .. ), and since the real ls -a command always shows . and .. at the top of the list, I have to add these manually and then sort everything with locale.strxfrm, so that they output the same as the real ls -a output. :)

implement-shell-tools/wc/wc.py
Geraldine-Edwards: Thanks again, I will add a comment to this along the lines of  "count lines by '\n'; add 1 if last line has no newline to match real wc command behavior"
:)

OracPrime (approved):
Good work, and thanks for the education on Python locales!

illicitonion:

PR #268, opened 2025-12-01
LONDON | SDC-NOV-2025| JESUS DEL MORAL | Sprint 4 | Feature / Implement Shell Tools Python 

delmorallopez:
Re-implement shell tools you have used (Used Python)

Each directory contains instructions for one shell tool to re-implement.

3 files changed in this PR cat (script-02.py), ls (script-02.py), ws (script-02.py)

### DIFF
--- a/implement-shell-tools/ls/script-02.py
+++ b/implement-shell-tools/ls/script-02.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import argparse
+
+
+def list_directory(path, show_all):
+    """List contents of a directory, respecting the -a and -1 options."""
+
+    try:
+        entries = os.listdir(path)
+    except FileNotFoundError:
+        print(f"ls: cannot access '{path}': No such file or directory", file=sys.stderr)
+        return
+
+    # If -a is not provided, hide dotfiles
+    if not show_all:
+        entries = [e for e in entries if not e.startswith(".")]
+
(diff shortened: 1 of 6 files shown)


implement-shell-tools/ls/script-02.py
```diff
@@ -0,0 +1,55 @@
...
+        dest="one_per_line",
+        action="store_true",
+        help="list one file per line",
+    )
+    parser.add_argument("path", nargs="?", default=".", help="directory to list")
+
+    args = parser.parse_args()
+
+    # Only -1 is supported, but it's always required in this assignment
+    if not args.one_per_line:
```
LonMcGregor: Why is this required? is there a reason i should not output the files on the same line?

LonMcGregor (commented):
Good start on these tasks, some things to consider:

Can you think of any problems that might arise from the filenames you have given your scripts?

See also my comment in the files. Mostly these implementations are very good and neat, well done

implement-shell-tools/ls/script-02.py
LonMcGregor: If you re-read the readme for the ls task, can you see where it is asking for -1 to be implemented? You need to be able to test with it on and off. Right now, every time it runs it prints each file on a new line, so even though you removed the option, how will you test if this option is implemented?

LonMcGregor (commented):
Great work,this task is complete now

illicitonion:

PR #270, opened 2025-12-01
London | 25-SDC-Nov | Aida Eslamimoghadam | Sprint 4 | Implement Shell tools in python 

aydaeslami:

This PR includes my solutions to simulate several shell tools in Python.
I used the argparse module to re-implement them.

### DIFF
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,30 +1,44 @@
 import argparse
-parser = argparse.ArgumentParser(
-    prog="my-cat",
-    description="Simple cat clone with -n and -b options",
-)
 
-counterNumber = 1
-parser.add_argument("-n", action="store_true", help="number all lines")
-parser.add_argument("-b", action="store_true", help="The character to search for" )
-parser.add_argument("path", nargs="+", help="The file to search")
 
-args = parser.parse_args()
+def run(args):
+    for file_path in args.path:
+        counter_number = 1
 
-for file_path in args.path:
-    with open(file_path, "r") as f:
-        content = f.read()
-        arrText = content.split("\n")
+        with open(file_path, "r") as f:
+            lines = f.readlines()
 
         if args.b:
-            for i in range(len(arrText )):
-                if arrText[i].strip() != "":
-                    print(counterNumber,arrText[i])
-                    counterNumber += 1
+            for line in lines:
+                if line.strip() != "":
+                    print(counter_number, line, end="")
+                    counter_number += 1
+                else:
+                    print(line, end="")
 
         elif args.n:
-          for i in range(len(arrText )):
-              print(counterNumber, arrText[i])
-              counterNumber += 1
+            for line in lines:
+                print(counter_number, line, end="")
+                counter_number += 1
+
         else:
-            print(content)
+            for line in lines:
+                print(line, end="")
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        prog="my-cat",
+        description="Simple cat clone with -n and -b options",
+    )
+
+    parser.add_argument("-n", action="store_true", help="number all lines")
+    parser.add_argument("-b", action="store_true", help="The character to search for")
+    parser.add_argument("path", nargs="+", help="The file to search")
+
+    args = parser.parse_args()
+    run(args)
+
+
+if __name__ == "__main__":
+    main()
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,7 +1,7 @@
 import argparse
 parser = argparse.ArgumentParser(
-    prog="count-containing-words",
-    description="Counts words in a file that contain a particular character",
+    prog="my-cat",
+    description="Simple cat clone with -n and -b options",
 )
 
 counterNumber = 1
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,0 +1,30 @@
+import argparse
+parser = argparse.ArgumentParser(
+    prog="count-containing-words",
+    description="Counts words in a file that contain a particular character",
+)
+
+counterNumber = 1
+parser.add_argument("-n", action="store_true", help="number all lines")
+parser.add_argument("-b", action="store_true", help="The character to search for" )
+parser.add_argument("path", nargs="+", help="The file to search")
+
+args = parser.parse_args()
+
+for file_path in args.path:
+    with open(file_path, "r") as f:
+        content = f.read()
+        arrText = content.split("\n")
+
+        if args.b:
+            for i in range(len(arrText )):
+                if arrText[i].strip() != "":
+                    print(counterNumber,arrText[i])
+                    counterNumber += 1
+
+        elif args.n:
+          for i in range(len(arrText )):
+              print(counterNumber, arrText[i])
+              counterNumber += 1
+        else:
+            print(content)
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -7,24 +7,14 @@ def run(args):
 
         with open(file_path, "r") as f:
             lines = f.readlines()
-
-        if args.b:
             for line in lines:
-                if line.strip() != "":
-                    print(counter_number, line, end="")
+                prefix = ""
+                if args.n or (args.b and line.strip() != ""):
+                    prefix = counter_number
                     counter_number += 1
-                else:
-                    print(line, end="")
+                print(f"{prefix} {line}", end="")
 
-        elif args.n:
-            for line in lines:
-                print(counter_number, line, end="")
-                counter_number += 1
-
-        else:
-            for line in lines:
-                print(line, end="")
-
+ 
 
 def main():
     parser = argparse.ArgumentParser(
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -14,7 +14,6 @@ def run(args):
                     counter_number += 1
                 print(f"{prefix} {line}", end="")
 
- 
 
 def main():
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,30 @@
...
+                if arrText[i].strip() != "":
+                    print(counterNumber,arrText[i])
+                    counterNumber += 1
+
+        elif args.n:
+          for i in range(len(arrText )):
+              print(counterNumber, arrText[i])
+              counterNumber += 1
+        else:
+            print(content)
```
DaryaShirokova: Could you please wrap you logic into a method and run it using `if __name__ == "__main__":` approach? Ideally the program should not contain global vars, and having methods improves readability.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,30 @@
...
+parser.add_argument("-n", action="store_true", help="number all lines")
+parser.add_argument("-b", action="store_true", help="The character to search for" )
+parser.add_argument("path", nargs="+", help="The file to search")
+
+args = parser.parse_args()
+
+for file_path in args.path:
+    with open(file_path, "r") as f:
+        content = f.read()
+        arrText = content.split("\n")
```
DaryaShirokova: There is a method in python to read file into an array of lines directly.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,30 @@
...
+
+        if args.b:
+            for i in range(len(arrText )):
```
DaryaShirokova: You can potentially use `enumerate` here.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,39 @@
...
+        print(file)
+
+
+parser = argparse.ArgumentParser(
+    prog="my-ls",
+    description="Simple ls clone with -a and -l options",
+)
+
+parser.add_argument("-a", action="store_true", help="include hidden files")
+parser.add_argument("-1", dest="one" ,action="store_true", help="use a long listing format")
```
DaryaShirokova: What do you mean by long listing format here?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,39 @@
+import argparse
+import os
+
+def show_unhidden_files(listDir):
```
DaryaShirokova: Can these two functions be merged into one using extra method param?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,66 @@
...
+lineCounter = 0
+wordCounter = 0
+charCounter = 0
+parser.add_argument("-l", action="store_true", help="count lines")
+parser.add_argument("-w", action="store_true", help="count words")
+parser.add_argument("-c", action="store_true", help="count characters")
+parser.add_argument("path", nargs="+", default=".", help="The file to count")
+args = parser.parse_args()
+
+if not args.l and not args.w and not args.c:
```
DaryaShirokova: Can you rework this code to have less duplication, by using methods?

DaryaShirokova (commented):
Great progress on this! Left a few comments here

implement-shell-tools/ls/ls.py
aydaeslami: Many thanks for your time and help.
I’ll update the help text to reflect that.

implement-shell-tools/ls/ls.py
aydaeslami: Thanks for the suggestion.
I merged them into a single function with an extra parameter.

implement-shell-tools/wc/wc.py
aydaeslami: Thank you for your feedback, I’ve refactored the code to reduce duplication.

implement-shell-tools/cat/cat.py
aydaeslami: Thanks for the suggestion. I’ve moved the logic into functions, added a main() entry point using if __name__ == "__main__": and removed global variables to improve readability.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,30 @@
+import argparse
+parser = argparse.ArgumentParser(
+    prog="my-cat",
+    description="Simple cat clone with -n and -b options",
+)
+
+counterNumber = 1
```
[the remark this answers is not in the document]
aydaeslami: Thanks for pointing that out. I’ll update the variable.

implement-shell-tools/cat/cat.py
aydaeslami: Thank you for the suggestion. I’ll use the built-in method to read the file into a list of lines directly.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,44 @@
...
+                    print(line, end="")
+
+        elif args.n:
+            for line in lines:
+                print(counter_number, line, end="")
+                counter_number += 1
+
+        else:
+            for line in lines:
+                print(line, end="")
```
illicitonion: This works, but often when doing this kind of "there are a few ways we may want to change things, but we're doing the same thing at its core", we try to structure code to emphasise what's the same/different, so something like:
```python
for line in lines:
    prefix = ""
    if args.n or (args.b and line.strip() != ""):
        prefix = counter_number
        counter_number += 1
    print(f"{prefix} {line}")
```

This makes clear that regardless of flags, we're doing something for each line, we're printing the line, and we _may_ print something different at the start.

It particularly helps when we end up with lots of different combinations, to think how each affects what we're doing, rather than having to handle them all separately.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+
+
+def show_files(files, show_hidden):
+    for file in files:
+        if show_hidden:
+            print(file)
+        else:
+            if not file.startswith("."):
+                print(file)
```
illicitonion: Same idea here:
```python
if show_hidden or not file.startswith("."):
    print(file)
```

This way if we change how we print things, we only need to change it in once place, not two.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,31 @@
...
+
+parser.add_argument("-a", action="store_true", help="include hidden files")
+parser.add_argument(
+    "-1", dest="one", action="store_true", help="list one entry per line"
+)
+parser.add_argument("path", nargs="?", default=".", help="The directory to list")
+args = parser.parse_args()
+
+
+fn = args.path
```
illicitonion: What does the variable name `fn` mean?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,31 @@
...
+listDir = os.listdir(fn)
+
+if fn != "":
```
illicitonion: When would `fn` be empty? From the naming here, it's pretty hard to work out what this means.

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,55 @@
...
+        print(lines, path)
+    elif args.w:
+        print(words, path)
+    elif args.c:
+        print(chars, path)
+    else:
+        print(lines, words, chars, path)
+
+if len(args.path) > 1 and not args.l and not args.w and not args.c:
+    print(lineCounter, wordCounter, charCounter, "total")
```
illicitonion: The real `wc` shows totals even if you specify flags, e.g. `wc -c /file/1 /file/2` will show a total row.

implement-shell-tools/ls/ls.py
aydaeslami: Thank you for the suggestion @illicitonion. That was a really valuable point. 
I have refactored the code to centralise the printing logic.
Many thanks.

implement-shell-tools/ls/ls.py
aydaeslami: You're right, Thank you for your feedback @illicitonion .
Since the `path` argument defaults to `"."`, the condition would never be false, so I have removed the unnecessary check.
I’ve also renamed the variable to make it clearer that it represents a directory path.

implement-shell-tools/wc/wc.py
aydaeslami: Thank you for your feedback, Daniel. I have refactored my code accordingly. please let me know if any further changes are needed. I really appreciate your suggestion.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,39 @@
...
+    prog="my-ls",
+    description="Simple ls clone with -a and -l options",
+)
+
+parser.add_argument("-a", action="store_true", help="include hidden files")
+parser.add_argument("-1", dest="one" ,action="store_true", help="use a long listing format")
+parser.add_argument("path", nargs="?", default=".", help="The directory to list")
+args = parser.parse_args()
+
+
```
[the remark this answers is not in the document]
aydaeslami: Thank you for the suggestion @DaryaShirokova. 
I’ve formatted the code using ruff to improve consistency and readability.

implement-shell-tools/cat/cat.py
aydaeslami: Thank you @illicitonion. that’s a really helpful suggestion. I’ve refactored the code to emphasise the shared logic and handle the differences through a prefix, as you described.

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,57 @@
...
+
+if len(args.path) > 1:
+    if args.l:
+        print(lineCounter, "total")
+    elif args.w:
+        print(wordCounter, "total")
+    elif args.c:
+        print(charCounter, "total")
+    else:
+        print(lineCounter, wordCounter, charCounter, "total")
```
illicitonion, line 57: The real `wc` doesn't treat these as exclusive flags - you can specify both `-w` and `-l` in the same command and it will show both.

We didn't list this as a requirement, so I'm going to mark this PR as complete, but you may want to try matching this behaviour if you're interested :)

illicitonion:

PR #277, opened 2025-12-03
London | 25-SDC-Nov | Jesus del Moral | Sprint 4 | Implement cowsay in Python 

delmorallopez:
Implement Cowsay Library in Python

Write a command line tool which uses the cowsay library to show user-supplied text, as said by a specified animal.

Used the cowsay library, and not write down the list of supported animals yourself, just fetch this information from the library.


LonMcGregor (commented):
Good start. Some feedback:

* It is generally not a good idea to be reading library files directly like this. What if the library updates and changes where they are stored?
* You are missing a requirements file, so I can't run this unless I already know which library needs to be installed

delmorallopez:
I’ve updated the script to use the cowsay public API instead of reading library files, and added a requirements.txt so the project is reproducible.

LonMcGregor (commented):
Good work, this is almost complete. When you run it, do you see it printing the message "None" after the output is finished? can you figure out why that is happening and remove it?

LonMcGregor:
Great work, this task is complete now

illicitonion:

PR #281, opened 2025-12-04
London | SDC-Nov-2025 | Ikenna Agulobi | Sprint 3| Implement shell tools

ike-agu:
## Learners, PR Template

Self checklist

## Changelist

This PR contains the completed implementations for the Sprint 3 exercises:
- cat command — file output with optional line numbering
- ls command — directory listing with -1 and -a support
- wc command — line, word, and character counting


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,44 @@
...
+
+const args = program.args; //Array all file paths
+
+//read flags user typed and return them as object.
+const opts = program.opts();
+
+let lineNumber = 1;
+
+// Loop over every filepath in args
+args.forEach(async (filepath) => {
```
eliza-an: Although this works it's not the most readable of implementations. Would your peer be able to read this easily and understand what is happening? Likely no! So it's super important our code is as readable as poossible, not necissarily as short as possible!
eg: 

for (const path of filepaths) {
    const filesContent = await fs.readFile(path, "utf-8");
    this will make it easier for everyone to understand

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,44 @@
...
+  const content = await fs.readFile(filepath, "utf8");
+  const lines = content.split("\n");
+
+  lines.forEach((line) => {
```
eliza-an: similarly here the arrow function unnecisarily complicates a simple bit of code. Using the example i gave you above, how would you rewrite this?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,80 @@
+import { program } from "commander";
+import { promises as fs } from "node:fs";
+import process from "node:process";
+
+program
+  .name("wc command")
```
eliza-an: this name is in a different format to your other names, not the end of the world, but consistency is always encouraged!

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,80 @@
...
+  .description("Implement wc command to count words and lines")
+  .option("-l,", "show number of lines only")
+  .option("-w,", "show number of words only")
+  .option("-c,", "show number of characters only")
+  .argument("[path...]", "The file path to process");
```
eliza-an: It is important to note that [] brackets mean the parameter is options and the <> brackets mean it is required. Remember this for my next comment below

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,80 @@
+import { program } from "commander";
+import { promises as fs } from "node:fs";
+import process from "node:process";
+
+program
+  .name("wc command")
+  .description("Implement wc command to count words and lines")
+  .option("-l,", "show number of lines only")
```
eliza-an: these commas in the brackets shouldnot exist! your code won't work then!

eliza-an (commented):
I left a few notes for you. This is a really god first try. There are some bits you need to change. Comments are great, but your code should be so readable that comments are unnecessary ( or at least as readable as can get!). If you need comments to understand your code, you need to simplify it so that you can read it easier! Well done overall. A good try, but some things need to be changed so your code works appropriately!

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,44 @@
...
+program
+  .name("display-file-content")
+  .description("Implement cat command with -n and -b flag support")
+  .option("-n, --number-all-lines", "Number every line in the file")
+  .option("-b, --number-non-empty-lines", "Number non empty lines in the file")
+  .argument("<paths...>", "File paths to process");
+
+program.parse(process.argv);
+
+const args = program.args; //Array all file paths
```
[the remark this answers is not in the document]
ike-agu: You're absolutely right. I've updated my code. Thanks for the feedback🙂

implement-shell-tools/cat/cat.js
ike-agu: Thanks for pointing this out and recommending  the for...of loop instead of the forEach loop. The code is much cleaner and easier to read after I implemented the fix.

implement-shell-tools/cat/cat.js
ike-agu: I used the for....of loop to refactor my code and it's much cleaner and readable. Thank you🙂

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,42 @@
...
+
+let paths = program.args; // array of paths user typed
+if (paths.length === 0) {
+  paths = ["."];
+}
+
+for (const directoryPath of paths) {
+  let entries;
+  try {
+    entries = await fs.readdir(directoryPath);
```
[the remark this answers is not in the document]
ike-agu: I agree, entries doesn’t say much about the code. I've updated the name to fileNames. Thanks!🙏🏽

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,42 @@
...
+  } catch (err) {
+    console.error(`ls: cannot access '${directoryPath}': ${err.message}`);
+    continue; // move to next path if this one fails
+  }
+
+  if (!opts.allFiles) {
+    entries = entries.filter((name) => !name.startsWith("."));
```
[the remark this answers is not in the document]
ike-agu: I’ve implemented the fix in a single line using filter.  The condition is now clearer and more readable. Thanks!

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,80 @@
...
+  .argument("[path...]", "The file path to process");
+
+program.parse(process.argv);
+
+const opts = program.opts();
+let paths = program.args;
+
+if (paths.length === 0) {
+  console.error("wc: no file specified");
+  process.exit(1);
```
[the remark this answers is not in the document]
ike-agu: Thank you for highlighting the error. I've implemented the fix so that if a path is not specified, it defaults to the current path.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,80 @@
...
+for (const filePath of paths) {
+  let content;
+  try {
+    content = await fs.readFile(filePath, "utf8");
+  } catch (err) {
+    console.error(`wc: cannot read file "${filePath}": ${err.message}`);
+    continue; // move on to next path
+  }
+
+  const lineCount = content.split("n").length;
```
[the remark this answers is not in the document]
ike-agu: I forgot the back slash  '\' which  means my code splits on letter n instead of splitting on a new line like so '\n'.  I've implemented the fix. Thanks🙏

implement-shell-tools/wc/wc.js
ike-agu: Thanks for pointing out that bug. I've updated my code  with the correct flags.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,80 @@
...
+  totals.words += wordCount;
+  totals.chars += charCount;
+
+  // Decide what to print depending on flags
+  if (!opts.l && !opts.w && !opts.c) {
+    console.log(`${lineCount} ${wordCount} ${charCount} ${filePath}`);
+    continue;
+  }
+
+  if (opts.l) {
```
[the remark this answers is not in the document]
ike-agu: I've implement a fix so that the formatting aligns with wc command. The code now works as expected. Thanks 🙏

ike-agu:
Hi @eliza-an , apologies for circling back to this PR after such a long time. I should have addressed the feedback earlier.
Thank you again for taking the time to review it and for the valuable feedback. 
I’ve now implemented all the requested changes.
When you have a moment, could you please take another look? Thank you!

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,42 @@
...
+const options = program.opts();
+
+let lineNumber = 1;
+
+for (const filepath of filepaths) {
+  const fileContent = await fs.readFile(filepath, "utf8");
+  const lines = fileContent.split("\n");
+
+  for (const line of lines) {
+    if (options.numberAllLines) {
```
illicitonion, line 25: This works, but something to think about for the future - here you're using if conditions to switch behaviours completely. But _most_ of the behaviours you want in all of the cases are actually the same (or at least, the same shape). So often we'll structure this code more like:

```js
let prefix = "";
if (options.numberAllLines || (options.numberNonEmptyLines && line.trim() !=== "") {
  prefix = `${lineNumber} `;
  lineNumber++;
}
console.log(`${prefix} ${line}`
```

This makes it more clear what's the _same_ in all of the cases, and what's _different_ between them. It also means that if we add more small changes (e.g. an option to put something at the end of the line) these compose well, rather than exploding into lots of separate options / ifs.

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,39 @@
...
+
+  try {
+    fileNames = await fs.readdir(directoryPath);
+  } catch (err) {
+    console.error(`ls: cannot access '${directoryPath}': ${err.message}`);
+    continue;
+  }
+
+  fileNames = fileNames.filter(
+    (file) => options.allFiles || !file.startsWith("."),
```
illicitonion, line 29: This is a really clear way of expressing the constraint! 🎉

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,63 @@
...
+
+  if (!opts.lines && !opts.words && !opts.chars) {
+    output.push(lineCount, wordCount, charCount);
+  } else {
+    if (opts.lines) output.push(lineCount);
+    if (opts.words) output.push(wordCount);
+    if (opts.chars) output.push(charCount);
+  }
+  console.log(`${output.join(" ")} ${filePath}`);
+}
```
illicitonion, line 50: This code is very similar to the code for accumulating totals. Extracting a function for outputting a line may be useful here (or may not - try it - it's not clear for this level of similarity/difference which would be more clear :))

illicitonion:

PR #285, opened 2025-12-05
Sheffield | 25-SDC-Nov | Hassan Osman | Sprint 4 | Implement shell tools (cat, ls, wc) in Python

HassanOHOsman:
Self checklist

This PR provides implementations of shell tools: cat, ls, and wc in Python.


implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,32 @@
...
+all_content = content
+
+if not args.a:
+    all_content = list(filter(lambda name: not name.startswith("."), content))
+
+for item in all_content:
+    if args.one:
+        print(item)
+    else:
+        print(item + " ")
```
OracPrime: Did you try this without the -1 argument? What output did you get and how did it compare to ls?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,32 @@
...
+parser.add_argument("path", help="The file path to read from")
+
+args = parser.parse_args()
+
+content = os.listdir(args.path)
+
+all_content = content
+
+if not args.a:
+    all_content = list(filter(lambda name: not name.startswith("."), content))
```
OracPrime: This lambda expression will work: can you think of a simpler method using a for expression?

OracPrime (commented):
A few bits of sorting out needed here.

implement-shell-tools/ls/ls.py
HassanOHOsman: > Did you try this without the -1 argument? What output did you get and how did it compare to ls?

Without -1 it lists every item in separate line, as if the -1 flag is added to the command line invocation. This is not the behaviour of ls (on its own without flags) where all items all listed on the same line.

HassanOHOsman: > If you try it with -a, how does it differ from the ls output?

1) same issue as the one above it.
2) the . and .. which refer to the current and parent dir(s) respectively, aren't shown in my implementation.

implement-shell-tools/ls/ls.py
HassanOHOsman: Done!

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,177 @@
+# import { program } from "commander";
```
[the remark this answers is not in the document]
HassanOHOsman: Agree! Used it as a guide for my python implementation and forgot to discard it.

implement-shell-tools/ls/ls.py
```diff
@@ -25,7 +25,7 @@
             if args.one:
                 print(item)
             else:
-                print(item, end=" ")
+                print(item, end="  ")
```
OracPrime: Why two spaces? Should it be \t? I don't actually know, but I think you should find out :)

implement-shell-tools/wc/wc.py
```diff
@@ -21,6 +21,10 @@ def counter(item):
     return {"lines": lines, "words": words, "characters": characters}
 
 
+def process_file(file_path):
+    global total_lines, total_words, total_characters, file_count
```
OracPrime: global is rarely a good word to see in source code... what's it doing here?

OracPrime (commented):
You've made it very hard to review - that in itself is a problem.  Don't commit individual file changes, commit self-consistent changes to your code.  I should be able to pull any commit and it should work.  I should be able to look at a single commit and see what the big change is.

At least one comment needs addressing - the double space in the ending.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,42 @@
...
+import os
+
+parser = argparse.ArgumentParser(
+    prog="list-files-in-directory",
+    description="List all files and directories in a directory",
+)
+
+parser.add_argument("-1", "--one", dest="one", help="Output one entry per line", action="store_true")
+parser.add_argument("-a", help="List all files & directories, including hidden ones", action="store_true")
+parser.add_argument("paths", nargs="*", default=["."], help="The file path to read from")
```
OracPrime: Arg is paths, comment is path.  That will confuse people.

implement-shell-tools/wc/wc.py
```diff
@@ -30,13 +30,13 @@ def process_file(file_path):
     stats = counter(content)
 
     if args.line:
-        print(f"{stats['lines']} {path}")
+        print(f"{stats['lines']} {file_path}}")
```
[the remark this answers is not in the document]
HassanOHOsman: My bad, I thought it frequent committing is the way. I will aim to group my commits moving forward. Thank you!

OracPrime (approved):
Nice work.  I think you're done with this one now!

illicitonion:

PR #287, opened 2025-12-07
Sheffield | 25-SDC-Nov | Sheida Shabankari | Sprint 5 |Laptop allocation

sheida-shab:

### DIFF
--- a/laptop_allocation.py
+++ b/laptop_allocation.py
@@ -28,12 +28,15 @@ laptops = [
     Laptop(id=2, manufacturer="Dell", model="XPS", screen_size_in_inches=15, operating_system=OperatingSystem.UBUNTU),
     Laptop(id=3, manufacturer="Dell", model="XPS", screen_size_in_inches=15, operating_system=OperatingSystem.UBUNTU),
     Laptop(id=4, manufacturer="Apple", model="macBook", screen_size_in_inches=13, operating_system=OperatingSystem.MACOS),
+    Laptop(id=5, manufacturer="Apple", model="macBook", screen_size_in_inches=13, operating_system=OperatingSystem.MACOS),
+    Laptop(id=6, manufacturer="Apple", model="macBook", screen_size_in_inches=13, operating_system=OperatingSystem.ARCH),
 ] 
 people = [
     Person(name="Imran", age=22, preferred_operating_system=[OperatingSystem.ARCH,OperatingSystem.UBUNTU]),
     Person(name="Eliza", age=34, preferred_operating_system=[OperatingSystem.ARCH,OperatingSystem.MACOS,OperatingSystem.UBUNTU]),
     Person(name="Leila", age=45, preferred_operating_system=[OperatingSystem.MACOS,OperatingSystem.UBUNTU,OperatingSystem.ARCH]),    
     Person(name="Mary", age=35, preferred_operating_system=[OperatingSystem.MACOS,OperatingSystem.ARCH]),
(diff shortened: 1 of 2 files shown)


laptop_allocation.py
```diff
@@ -0,0 +1,80 @@
...
+people = [
+    Person(name="Imran", age=22, preferred_operating_system=[OperatingSystem.ARCH,OperatingSystem.UBUNTU]),
+    Person(name="Eliza", age=34, preferred_operating_system=[OperatingSystem.ARCH,OperatingSystem.MACOS,OperatingSystem.UBUNTU]),
+    Person(name="Leila", age=45, preferred_operating_system=[OperatingSystem.MACOS,OperatingSystem.UBUNTU,OperatingSystem.ARCH]),    
+    Person(name="Mary", age=35, preferred_operating_system=[OperatingSystem.MACOS,OperatingSystem.ARCH]),
+    Person(name="Sara", age=28, preferred_operating_system=[OperatingSystem.MACOS])
+]
+
+# Global sadness counter
+sadness=0
```
OracPrime: Having a global variable is normally a bad idea.  This could just as easily be created inside the allocate_laptops function and returned as part of a multi-value return from the function.  As it is if you call the function twice the second time will include the sadness from the first time in its sadness score.

laptop_allocation.py
```diff
@@ -0,0 +1,80 @@
...
+    allocated_history : Dict[Person,Laptop] ={}
+    global sadness
+    for person in sorted_people_OS_count :
+     allocated_flag=False
+     for i in range(len(person.preferred_operating_system)) :
+          for laptop in laptops :
+              if person.preferred_operating_system[i] == laptop.operating_system :
+                  allocated_history[person.name]=laptop # assign laptop
+                  sadness += i  # increment sadness by preference index
+                  laptops.remove(laptop)
```
OracPrime: The laptops argument is passed to allocate_laptops by reference.  So this remove function is modifying the original list of laptops.  Now this may be intentional, in which case it should be made obvious somewhere.  Alternatively make a copy of the list at the start of the function and modify that inside the function.  I can see arguments for both approaches being desirable, but it would be good to make it obvious.

laptop_allocation.py
```diff
@@ -0,0 +1,80 @@
...
+                  allocated_flag=True
+                  break
+          if allocated_flag : 
+            break     
+          
+     if not allocated_flag :   # assign any remaining laptop if preferred OS not found
+        allocated_history[person.name]=laptops[0]
```
OracPrime: Both here and above you're using person.name as they key in the dictionary.  On line 53 you declared the key as Person.  One or the other needs correcting.

OracPrime (commented):
Generally solid code, but a couple of bits need tidying up or clarifying.

laptop_allocation.py
sheida-shab: Thanks for your feedback.
I fixed it.

sheida-shab:
Hi David,

I just wanted to let you know that I’ve addressed all the points you mentioned in your previous comments. I noticed that you reviewed the PR again, but there weren’t any new comments added, so I’m not sure whether the PR is now considered complete or if there are still any issues to fix.

If possible, could you please take another look and let me know the current status?

Thank you very much for your time and support.

OracPrime:
Variable naming and readability are really important code quality issues, which is why I flagged it for one final fix

illicitonion:

PR #297, opened 2025-12-10
  London | 25-SDC-Nov | Jesus del Moral | Sprint 5 | Implement Laptop Allocation

delmorallopez:
Laptop allocation app

Every person should be allocated exactly one laptop.

If we define “sadness” as the number of places down in someone’s ranking the operating system the ended up with (i.e. if your preferences were [UBUNTU, ARCH, MACOS] and you were allocated a MACOS machine your sadness would be 2), we want to minimise the total sadness of all people. If we allocate someone a laptop with an operating system not in their preferred list, treat them as having a sadness of 100

### DIFF
--- a/laptop-allocation.py
+++ b/laptop-allocation.py
@@ -1,0 +1,81 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Dict, Tuple
+
+class OperatingSystem(Enum):
+    MACOS = "macOS"
+    ARCH = "Arch Linux"
+    UBUNTU = "Ubuntu"
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: int
+    preferred_operating_system: Tuple[OperatingSystem, ...]
+
+@dataclass(frozen=True)
+class Laptop:
+    id: int
+    manufacturer: str
+    model: str
+    screen_size_in_inches: float
+    operating_system: OperatingSystem
+
+
+# Assign each person the best available laptop according to their preference list
+# Minimise sadness locally (not globally optimal, but simple and valid for this exercise)
+# If no matching OS is available → they get the “least bad” remaining laptop (sadness = 100)
+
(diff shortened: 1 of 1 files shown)


laptop-allocation.py
```diff
@@ -0,0 +1,81 @@
...
+class OperatingSystem(Enum):
+    MACOS = "macOS"
+    ARCH = "Arch Linux"
+    UBUNTU = "Ubuntu"
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: int
+    preferred_operating_system: Tuple[OperatingSystem, ...]
```
OracPrime: 1) Why a Tuple, not a List?
2) If it is a collection, it should have a pluralised name (preferred_operating_systems)

laptop-allocation.py
```diff
@@ -0,0 +1,81 @@
...
+def sadness(person: Person, laptop: Laptop) -> int:
+    # Return sadness score for assigning this laptop to this person
+    if laptop.operating_system in person.preferred_operating_system:
+        return person.preferred_operating_system.index(laptop.operating_system)
+    return 100
+
+
+def allocate_laptops(people: List[Person], laptops: List[Laptop]) -> Dict[Person, Laptop]:
+    if len(people) != len(laptops):
+        raise ValueError("Number of people must equal number of laptops.")
```
OracPrime: Why does the number of people need to equal the number of laptops?

OracPrime (commented):
Couple of minor points, one bug, but generally good code, well done.

laptop-allocation.py
delmorallopez: 1.- I used Tuple instead of List to signal immutability and a fixed relationship, not something that will be modified. Items should not change. 
2.- I have changed the name "preferred_operating_systems"

illicitonion:

PR #307, opened 2025-12-18
London | 25-SDC-Nov | Emiliano Uruena | Sprint 5 |  Laptops Allocation

Emilianouz:
Self checklist

## Changelist

Implementing Laptop allocation exercise.

### DIFF
--- a/laptop-allocation.py
+++ b/laptop-allocation.py
@@ -1,0 +1,62 @@
+#If we define "sadness" as the number of places down in someone's ranking 
+#the operating system the ended up with (i.e. if your preferences were 
+#[UBUNTU, ARCH, MACOS] and you were allocated a MACOS machine your sadness would be 2), 
+#  0       1      2      100
+#we want to minimize the total sadness of all people. If we allocate someone a laptop 
+#with an operating system not in their preferred list, treat them as having a sadness of 100.
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Dict, Optional
+
+
+class OperatingSystem(Enum):
+    MACOS = "macOS"
+    ARCH = "Arch Linux"
+    UBUNTU = "Ubuntu"
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: int
+    preferred_operating_system: List[OperatingSystem]
+    assigned_laptop: Optional[int] = None
+
+@dataclass(frozen=True)
+class Laptop:
+    id: int
+    manufacturer: str
+    model: str
+    screen_size_in_inches: float
+    operating_system: OperatingSystem
+    assigned: bool = False
+
+def allocate_laptops(people: List[Person], laptops: List[Laptop]) -> Dict[str, int]:
+    allocation = {}
+    for person in people:
+        for os in person.preferred_operating_system:
+            for laptop in laptops:
+                if (laptop.operating_system == os and not laptop.assigned):
+                    allocation[person.name] = laptop.id
+                    assigned = True
+                    break
+            if assigned:
+                break
+    return allocation
+
+people = [
+    Person(name="Imran", age=22, preferred_operating_system=[OperatingSystem.UBUNTU,OperatingSystem.ARCH,OperatingSystem.MACOS]),
+    Person(name="Eliza", age=34, preferred_operating_system=[OperatingSystem.ARCH]),
+    Person(name="Maria", age=38, preferred_operating_system=[OperatingSystem.MACOS,OperatingSystem.ARCH]),
+]
+
+laptops = [
+    Laptop(id=1, manufacturer="Dell", model="XPS", screen_size_in_inches=13, operating_system=OperatingSystem.ARCH),
+    Laptop(id=2, manufacturer="Dell", model="XPS", screen_size_in_inches=15, operating_system=OperatingSystem.UBUNTU),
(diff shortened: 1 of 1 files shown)


number-systems/README.md
```diff
@@ -5,61 +5,62 @@ Do not convert any binary numbers to decimal when solving a question unless the
 The goal of these exercises is for you to gain an intuition for binary numbers. Using tools to solve the problems defeats the point.
 
 Convert the decimal number 14 to binary.
-Answer:
```
OracPrime: What's this file doing in the laptops-allocation task?

laptop-allocation.py
```diff
@@ -0,0 +1,62 @@
...
+    screen_size_in_inches: float
+    operating_system: OperatingSystem
+    assigned: bool = False
+
+def allocate_laptops(people: List[Person], laptops: List[Laptop]) -> Dict[str, int]:
+    allocation = {}
+    for person in people:
+        for os in person.preferred_operating_system:
+            for laptop in laptops:
+                if (laptop.operating_system == os and not laptop.assigned):
```
OracPrime: Have you run this code and checked the output?  Can you paste a copy as a comment so I can see it has worked?  Thanks

Emilianouz:
README file no necessary. It was removed.

laptop-allocation.py
```diff
@@ -0,0 +1,62 @@
...
+                    allocation[person.name] = laptop.id
+                    assigned = True
```
SlideGauge: Does this assigned variable prevent us from selecting the same laptop several times?

laptop-allocation.py
```diff
@@ -0,0 +1,62 @@
...
+                    break
+            if assigned:
```
SlideGauge: what happens if we did not found the laptop after the first loop
```for laptop in laptops```?

Comment (SlideGauge) [inline laptop-allocation.py]: if assigned was set to True, what happens after iterating the first os?

File: laptop-allocation.py [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+@dataclass(frozen=True)
+class Laptop:
+    id: int
+    manufacturer: str
+    model: str
+    screen_size_in_inches: float
+    operating_system: OperatingSystem
+    assigned: bool = False
+
+def allocate_laptops(people: List[Person], laptops: List[Laptop]) -> Dict[str, int]:
```
SlideGauge: The algorithm belongs to the type of algorithms called "Greedy", i.e. on each step the algorithm takes the most suitable choice seen **at this moment**, not **globally**.
Consider:
Imran wants UBUNTU, ARCH
Eliza wants UBUNTU

The algorithm will give UBUNTU to Imran and nothing to Eliza, leading to sadness 100.

But we can invent different algorithm: if we distribute UBUNTU to Eliza and ARCH to Imran, we will have sadness 1. This algorithm will be slower, but our task is not to create the quickest algorithm, but minimizing the sadness.

Hint: this task requires analyzing of all possible permutations of laptops distribution.

laptop-allocation.py
```diff
@@ -0,0 +1,62 @@
...
+    MACOS = "macOS"
+    ARCH = "Arch Linux"
+    UBUNTU = "Ubuntu"
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: int
+    preferred_operating_system: List[OperatingSystem]
+    assigned_laptop: Optional[int] = None
```
SlideGauge, line 23: Since the dataclass is frozen, we won't be able to change the value of this variable

SlideGauge:
Made review, interesting approach  but there are some errors in the implementation itself worthy fixing + the algorithm won't find the most minimal possible configuration

OracPrime:
I'll leave @SlideGauge to carry on with this since he's most recent reviewer

laptop-allocation.py
```diff
@@ -0,0 +1,96 @@
...
+    operating_system: OperatingSystem
+    assigned: bool = False
+
+# sadnes calculator
+def calculate_sadness(person: Person, laptop: Laptop) -> int:
+    if laptop.operating_system in person.preferred_operating_system:
+        return person.preferred_operating_system.index(laptop.operating_system)
+    return 100
+
+def allocate_laptops(people: List[Person], laptops: List[Laptop]) -> Dict[str, int]:
```
SlideGauge: Really nice, yes, backtrack will solve the task.

SlideGauge: One thing to take into account: 
the task demands us for the function to have this signature:
```def allocate_laptops(people: List[Person], laptops: List[Laptop]) -> Dict[Person, Laptop]:```
Note, we return Dict of  Person->Laptop, not str->int. It is valuable to take into account, as if we just return str->int, we won't be able to distinguish to people with the same name.

SlideGauge (commented):
Overall really good solution, please change the returning type of the function and I will mark the review as complete.

laptop-allocation.py
Emilianouz: Right. We should return Dict[Person, Laptop] so people with the same name are handled correctly.

Emilianouz:
@SlideGauge , I changed it to return type of the function: Person and Laptop objects. Thank you

SlideGauge:
Very nice, thank you, accepted

illicitonion:

PR #310, opened 2025-12-20
Glasgow | 25-SDC-Nov | Nataliia Volkova | Sprint 5 | Prep Exercises

Nataliia74:
## Learners, PR Template

Self checklist

## Changelist
01 Prep Exercises issue 178


dataclasses_ex.py
LonMcGregor: Can you describe the difference and any dis/advantage of using a class method vs a function?

familytree.py
```diff
@@ -0,0 +1,20 @@
...
+
+fatma = Person(name="Fatma", children=[], age=17)
+aisha = Person(name="Aisha", children=[], age=25)
+
+imran = Person(name="Imran", children=[fatma, aisha], age=51)
+
+def print_family_tree(person: Person) -> None:
+    print(person.name, f"({person.age})")
+    for child in person.children:
+        print(f"- {child.name} ({child.age})")
```
LonMcGregor: Can this print grandchildren recursively?

LonMcGregor (commented):
Very good work on these tasks, there are some areas where you could improve further

Also,consider if it is a good idea to have all these files in the main directory for this sprint

LonMcGregor:
Good improvements. I think there is still an issue about validating the OSes. You print them out now, which is good, but the program still immediately exits if you type them wrong. Can you get it working like the other inputs where it loops until the input is valid?

Also, did you see my comment asking about the impact of having the prep files all in the main directory rather than a prep specific folder?

illicitonion:

PR #324, opened 2026-02-07
Glasgow | 25-SDC-NOV | Katarzyna Kazimierczuk | Sprint 3 | Implement Shell Tools

katarzynakaz:
Self checklist

## Changelist

js tools.


implement-shell-tools/cat/index.js
```diff
@@ -0,0 +1,87 @@
...
+            
+        })
+    }
+
+}
+
+
+const argv = process.argv.slice(2);
+
+switch (argv[0]) {
```
SlideGauge: What is here is "pattern matching" of the cases from the task. Whilst it work for these cases, this is unfortunately not the way how production software is written.
Several moments:
1. functions are referring to some "cases". It is not clear what they mean. Instead, function naming should reflect what the function does: output lines, output lines with numbering etc
2. Hint: the code in all three functions has a lot of similarities: we read list of files, traverse the list one by one, apply some rule to the line (or no rule at all) and then output the line to console. Can we generalise this somehow, like have a function whose responsibility is to traverse the input list, then the function which reads the file and applies the rules? Try it. (hint: maybe usage of  functions as objects passed into initial function may help)

implement-shell-tools/cat/notes.js
```diff
@@ -0,0 +1,46 @@
+import process from "node:process";
+import { promises as fs } from "node:fs";
+// const fs = promises
```
SlideGauge: This file looks like unnecessary

implement-shell-tools/cat/index.js
```diff
@@ -0,0 +1,87 @@
+import process from "node:process";
+import { promises as fs } from "node:fs";
+
+
+//one for 1 and 3
+async function printOneOrMore(listOfFiles) {
```
SlideGauge: Nice that all functions are generalised to work with a list, instead (one function for single file, another function is for a list)

implement-shell-tools/cat/index.js
```diff
@@ -0,0 +1,87 @@
...
+//     })
+// }
+    
+// }
+
+//one instead
+//  2 * `cat -n sample-files/1.txt`
+// 4 * `cat -n sample-files/*.txt`
+
+async function caseTwoAndFour(listOfFiles) {
```
SlideGauge: Overall, we do not need async/await pattern in the cat-like utility. We just sequentially read file line by line and at the same time the utility is a console utility

implement-shell-tools/ls/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+
+const argv = process.argv.slice(2);
+
+if (argv.includes('-a')) {
+    await showAllInSampleFiles();
+} else if (argv.includes('sample-files')) {
+    await showVisibleInSampleFiles();
+} else {
+    await showCurrentDir();
```
SlideGauge: There is no such a function in this file

implement-shell-tools/cat/index.js
```diff
@@ -0,0 +1,87 @@
...
+    for (const file of listOfFiles) {
```
SlideGauge: cat -n with multiple files should continue line numbering across files — i.e., if file 1 has 3 lines, file 2 starts at line 4. Currently numbering restarts at 1 per file. You'd need to track a counter outside the for loop.

implement-shell-tools/wc/index.js
```diff
@@ -0,0 +1,77 @@
...
+// const countOfWordsContainingEs = content
+//   .split(" ")
+//   .filter((word) => word.includes("e"))
+//   .length;
+// console.log(countOfWordsContainingEs);
+
+function countHelper(inputFiles) {
+    return {
+        lines: inputFiles.split('\n').length - 1,
+        words: inputFiles.split(' ').filter(w => w !== "").length,
```
SlideGauge: It means we split only according to space, but will ignore for example newline character or other whitespace characters. Could you fix it please?

implement-shell-tools/wc/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+//from coursework
+// const content = await fs.readFile(path, "utf-8");
+// const countOfWordsContainingEs = content
+//   .split(" ")
+//   .filter((word) => word.includes("e"))
+//   .length;
+// console.log(countOfWordsContainingEs);
+
+function countHelper(inputFiles) {
```
SlideGauge: countHelper is not very clear name, could you rename so naming reflects the purpose of the function?

SlideGauge (commented):
General note: I notced async/await being used here, I have doubts in necessity of it in this context: we're doing a console utility

katarzynakaz:
Thank you for the feedback. I was folllowing the coursework in Sp 4 with the use of async / await but I see your point. 
Changes implemented :)

implement-shell-tools/ls/index.js
```diff
@@ -0,0 +1,41 @@
...
+	const listOfFiles = await fs.readdir(directory);
+
+	for (const eachFile of listOfFiles) {
+		if (eachFile[0] !== ".") console.log(eachFile);
+	}
+}
+
+// `ls -1 sample-files`
+async function showVisibleInSampleFiles() {
+	const listOfFiles = await fs.readdir("sample-files");
```
illicitonion, line 15: We generally don't want to hard-code that our program only works with specific directories - can you think how to make this more general? And how to make it share code with the other two functions?

implement-shell-tools/wc/index.js
```diff
@@ -0,0 +1,74 @@
...
+		console.log(`${counts.lines} ${counts.words} ${counts.bytes} ${file}`);
+	}
+}
+
+const argv = process.argv.slice(2);
+const files = argv.filter((arg) => !arg.startsWith("-"));
+
+if (argv.includes("-l")) {
+	await countLines(files);
+} else if (argv.includes("-w")) {
```
illicitonion, line 68: The real `wc` also supports passing multiple of the flags at once (e.g. `wc -w -l /some/file`), and does something a bit different if you pass multiple files (`wc -w -c /some/file /some/other/file`) - can you see if you can match those behaviours?

implement-shell-tools/wc/index.js
```diff
@@ -0,0 +1,74 @@
...
+	return {
+		lines: inputFiles.split("\n").length - 1,
+		words: inputFiles.split(/\s+/).filter((w) => w !== "").length,
+		bytes: inputFiles.length,
+	};
+}
+
+// * `wc -l sample-files/3.txt`
+// * `wc -l sample-files/*`
+async function countLines(listOfFiles) {
```
illicitonion, line 22: These functions are quite repetitive with each other.

implement-shell-tools/wc/index.js
```diff
@@ -0,0 +1,74 @@
+import process from "node:process";
+import { promises as fs } from "node:fs";
+
+//from coursework
```
illicitonion, line 4: Please remove the assorted commented out code :)

implement-shell-tools/cat/index.js
```diff
@@ -0,0 +1,50 @@
...
+		if (!flag) {
+			console.log(file);
+		} else if (flag === "-n") {
+			console.log(`${countingOnlyFullLines} ${file}`);
+			countingOnlyFullLines += 1;
+		} else if (flag === "-b") {
+			if (file === "") {
+				console.log("");
+			} else {
+				console.log(`${countingOnlyFullLines} ${file}`);
```
illicitonion, line 40: These ifs are quite repetitive, which has a couple of issues:
1. When reading the code, it's hard to see which parts are the same across the branches, and what parts are different.
2. In the future if we add new concerns (e.g. a "colour every other line" flag), we'll need to implement that across each of the branches.

We often solve this by separating "the differences" from "what's the same", e.g.

```js
// Make a variable up-front:
let prefix = "";
// Conditionally change things
if (something) {
  prefix = "hello";
} else if (somethingElse) {
  prefix = "goodbye";
}
console.log(`${prefix}${line}`);
```

So that the `if`s at the start make clear how things are different, and what comes after shows what's the same/different

illicitonion (commented):
A lot of the pieces here are looking good, but the solutions often aren't very general - can you take a look at these comments? Thanks!

implement-shell-tools/.DS_Store
illicitonion, line 1: Please can you remove these `.DS_Store` files from your branch? Thanks!

PR #326, opened 2026-02-09
Glasgow | 25-SDC-NOV | Katarzyna Kazimierczuk | Sprint 5 | prep

katarzynakaz:
Self checklist

prep tasks


prep/10.py
```diff
@@ -0,0 +1,60 @@
...
+    manufacturer: str
+    model: str
+    screen_size_in_inches: float
+    operating_system: str
+
+
+def find_possible_laptops(laptops: List[Laptop], person: Person) -> List[Laptop]:
+    possible_laptops = []
+    for laptop in laptops:
+        if laptop.operating_system == person.preferred_operating_systems:
```
SlideGauge: Could you doublecheck this line? (Hint: check types of operating_system and preferred_operating_systems. What will happen if we check vars of these types between each other? Is that what we intend to happen here?)

implement-shell-tools/cat/notes.js
```diff
@@ -0,0 +1,38 @@
+import process from "node:process";
```
SlideGauge: Looks like this file is not related to the Sprint 5

implement-shell-tools/ls/index.js
```diff
@@ -0,0 +1,50 @@
+// fs.readdir
```
SlideGauge: Looks like this file is not related to Sprint 5

prep/notes.md
```diff
@@ -0,0 +1,39 @@
+
```
SlideGauge: This is really good you're practising your ideas in a separate file. At the same time,  this file is not related into any particular task so it is a bit hard for me to review it. Could you doublecheck the necessity of this file in the PR?

prep/11.py
```diff
@@ -0,0 +1,109 @@
...
+    screen_size_in_inches: float
+    operating_system: OperatingSystem
+
+def create_a_new_person() -> Person:
+    add_name = input("What is your name? \n")
+    if not add_name or add_name == "" or len(add_name) > 50:
+        print("Enter a valid name.", file=sys.stderr)
+        sys.exit(1)
+
+    add_age = int(input("What is your age? \n"))
```
SlideGauge: What will happen if I enter not a number but something else?

prep/8_laptop_program.py
```diff
@@ -0,0 +1,109 @@
...
+
+def find_possible_laptops(laptops: List[Laptop], person: Person) -> List[Laptop]:
+    possible_laptops = []
+    other_laptops = []
+    for laptop in laptops:
+        if laptop.operating_system == person.preferred_operating_system:
+            possible_laptops.append(laptop)
+        else:
+            other_laptops.append(laptop)
+    return possible_laptops, other_laptops
```
SlideGauge, line 77: Could you double check what would be the type of an
```a, b```
expression?
Is it really a List, as your type hinting of the find_possible_laptops says?

File: prep/8_laptop_program.py [hunk 7]
```diff
@@ -0,0 +1,109 @@
...
+
+def compare_counts_of_poss_and_others(possible_laptops: List[Laptop], other_laptops: List[Laptop], person: Person) -> None:
+    # counter to get obj
+    other_laptops_counted_as_obj = Counter(laptop.operating_system for laptop in other_laptops)
+    # and here turn to tuple os+count
+    most_common_laptop = other_laptops_counted_as_obj.most_common(1)[0]
```
SlideGauge, line 86: in theory other_laptops can be empty list, thus Counter most_common will return empty collection either. What will happen if we try to access zero index of an empty list? Can we avoid this?

prep/4_Person_class_exercise.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, name: str, dob: date, preferred_operating_system: str):
+        self.name = name
+        self.dob = dob
+        self.preferred_operating_system = preferred_operating_system
+        
+
+    def is_adult(self):
+        today = date.today()
+        age = date.today().year - self.dob.year
+        if (today.month, today.day) < (self.dob.month, self.dob.day):
```
SlideGauge, line 20: Nice, you're handling the case of partial year.

sprint 5/laptop.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@dataclass(frozen=True)
+class Laptop:
+    id: int
+    manufacturer: str
+    model: str
+    screen_size_in_inches: float
+    operating_system: OperatingSystem
+Write a function with this signature:
```
SlideGauge: Do we want this task to be implemented? I see empty functions here

SlideGauge (commented):
Could you please rearrange/rename prep tasks (instead of 1/2 etc) to the way they're named in https://sdc.codeyourfuture.io/tools/sprints/5/prep/#type-guided-refactorings
?
Otherwise it is a bit hard to deduce which task solved what.

SlideGauge:
Overall decent work, but some adjustments are needed for clarity.

sprint 5/laptop.py
katarzynakaz: aplogies, the completed task is in a separate pr

prep/10.py
katarzynakaz: thank you, oh yes it was a comparison of str to a list of str

prep/11.py
katarzynakaz: good point, a crash would happen. issue fixed

prep/8_laptop_program.py
katarzynakaz: thank you, I updated type hint to match the return of 2 lists

prep/8_laptop_program.py
katarzynakaz: empty list check added

katarzynakaz:
Thank you for the feedback. The files are renamed to match the exercises with a corresponding page of prep and the issues are fixed.

.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+DS_Store
```
SlideGauge: Tiny caveat: you need to name the file with the dot in front:
```.DS_Store```
Also note that fixing the gitignore doesn't retroactively untrack files already committed — you still need to remove them manually.

But good job in trying to fix the DS_Store issue by adding gitignore

**Review by SlideGauge — COMMENTED** [review]
Please, remove remaining .DS_Store files

File: prep/8 laptop program.py [hunk 11]
```diff
@@ -0,0 +1,116 @@
...
+    for laptop in laptops:
+        if laptop.operating_system == person.preferred_operating_system:
+            possible_laptops.append(laptop)
+        else:
+            other_laptops.append(laptop)
+    return possible_laptops, other_laptops
+
+def compare_counts_of_poss_and_others(possible_laptops: List[Laptop], other_laptops: List[Laptop], person: Person) -> None:
+    if len(other_laptops) == 0:
+        print("The os you chose has the biggest availability.)")
```
SlideGauge: Tiiiny moment: looks like .)" was placed just by accident

prep/8 laptop program.py
```diff
@@ -0,0 +1,116 @@
...
+        return
+    # counter to get obj
+    other_laptops_counted_as_obj = Counter(laptop.operating_system for laptop in other_laptops)
+    # and here turn to tuple os+count
+    most_common_laptop = other_laptops_counted_as_obj.most_common(1)[0]
+    
+    if len(possible_laptops) < most_common_laptop[1]:
+        # print(f"There are more laptops available with {str(most_common_laptop[0])} than your preferred OS {person.preferred_operating_system.value}. Consider accepting that OS to increase your chances of getting a laptop.")
```
SlideGauge: Remnants of debugging. Should be avoided for production code.

SlideGauge (commented):
Thanks a lot for renaming the filenames, now much clearer.
A small note here (it is not a problem in this case, but in future may be):
your .py files have spaces in their name which will produce problems if we try to import these files as modules. Whilst it is not a problem in this concrete sprint, in future this should be avoided (use underscores, this is a typical pattern)

SlideGauge (commented):
Also, could you place trailing newline in files. It hints you with 

Image
symbol that this is missing

SlideGauge:
Good job, most of my comments are about some code organization/filesystem organization, which worthy to remember to maintain. From the point of view of tasks themselves, answers are correct.
Please, adjust the things I mentioned and I'll accept it.

.gitignore
katarzynakaz: thank you!

prep/8 laptop program.py
katarzynakaz: thank you, both issues fixed

katarzynakaz:
Thank you very much, listed changes implemented.

SlideGauge (commented):
Nice job, review complete

illicitonion:

PR #331, opened 2026-02-22
Glasgow | 25-SDC-NOV | Katarzyna Kazimierczuk | Sprint 4 | Python shell tools

katarzynakaz:
Self checklist

## Changelist

Cat, wc and ls implemented in pyton.


implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,44 @@
+import sys
+from pathlib import Path
+
+def cleanInput(listOfFiles):
+    cleanLinesArr = []
+
+    for file in listOfFiles:
+        grabbedText = Path(file).read_text(encoding="utf-8") 
+        splitLines = grabbedText.split("\n")
```
OracPrime: There's a subtle bug here I think.  Consider what happens if the file ends with a line which ends with a newline.
Or to put it another way, what does `"Hello\n".split("\n")` return?

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,44 @@
...
+restIsFiles = []
+
+if len(args) > 0 and args[0] and args[0][0] == "-":
+    flag = args[0]
+    restIsFiles = args[1:]
+else:
+    flag = None
+    restIsFiles = args
+
+def takeSpecifiedAction(cleanLinesArr, flag):
```
OracPrime, line 14: It's bad form, normally, to hide functions inside the top-level code of a python file.  This should be before line 14.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,44 @@
...
+    countingOnlyFullLines = 1
+
+    for file in cleanLinesArr:
```
OracPrime: "file" isn't a file here, it's a line.  You need to rename the variable

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,44 @@
...
+            print(f"{countingOnlyFullLines} {file}")
+            countingOnlyFullLines += 1
+        elif flag == "-b":
+            if file == "":
+                print("")
+            else:
+                print(f"{countingOnlyFullLines} {file}")
+                countingOnlyFullLines += 1
+        else:
+            print("incorrect flag")
```
OracPrime: You should validate the flag outside the loop.  At the moment if the flag is invalid you'll print this message for every line of every file.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,34 @@
...
+def showAllFilesInDir(directory):
+    listOfFiles = os.listdir(directory)
+
+    for eachFile in listOfFiles:
+        if eachFile[0] != ".":
+            print(eachFile)
+
+# `ls -1 sample-files`
+def showVisibleInSampleFiles():
+    listOfFiles = os.listdir("sample-files")
```
OracPrime: Why is this path hard-coded here?

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,34 @@
...
+
+    for eachFile in listOfFiles:
+        print(eachFile)
+
+argv = sys.argv[1:]
+
+if "-a" in argv:
+    showAllInSampleFiles()
+elif "sample-files" in argv:
```
OracPrime: No - you shouldn't be worrying about sample-files as a special case.  You should take the arguments from the command line!  I suspect if you look back at this file now you have more experience you'll be reaching for the keyboard!

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,53 @@
+import sys
+from pathlib import Path
+
+def calculateCounts(inputFiles):
+    return {
```
OracPrime: This is counting lines, words and bytes every time, which is inefficient.  You should pass in a flag as to what you want, and just do a single calculation

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,53 @@
...
+        "lines": len(inputFiles.split("\n")) - 1,
```
OracPrime: split("\n") has the same bug discussed in cat.py

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,53 @@
...
+def countWords(listOfFiles):
+    for file in listOfFiles:
+        content = Path(file).read_text(encoding="utf-8")
+
+        # const wordsCounted = content.split(" ").filter(word => word !== "").length;
+        counts = calculateCounts(content)
+        print(f"{counts['words']} {file}")
+
+# * `wc -c sample-files/3.txt`
+def countBytes(listOfFiles):
```
OracPrime: These three functions could easily be combined into a single function with a flag.

number-systems/README.md
```diff
@@ -5,61 +5,61 @@ Do not convert any binary numbers to decimal when solving a question unless the
 The goal of these exercises is for you to gain an intuition for binary numbers. Using tools to solve the problems defeats the point.
 
 Convert the decimal number 14 to binary.
-Answer:
+Answer:1110
```
OracPrime: Isn't this file left over from a different sprint?  And we need .gitignore for .DS_Store

OracPrime (changes requested):
Some significant cleaning up needed on this one.

OracPrime:
I also don't understand why this has the "NotCoursework" tag - it's in the backlog

number-systems/README.md
```diff
@@ -5,61 +5,41 @@ Do not convert any binary numbers to decimal when solving a question unless the
 The goal of these exercises is for you to gain an intuition for binary numbers. Using tools to solve the problems defeats the point.
 
 Convert the decimal number 14 to binary.
-Answer:
```
OracPrime, line 8: This file is not part of this sprint.  You need to do a git restore of the file to get it back to the original state

OracPrime (changes requested):
One more file to remove

number-systems/README.md
katarzynakaz: issue fixed

OracPrime: Github doesn't agree.  It is showing differences.  If you do a restore to the start point of thie PR then this file shouldn't be showing up in the PR list at all.

OracPrime (changes requested):
Not quite there yet.

number-systems/README.md
katarzynakaz: thank you, that's now done

OracPrime (approved):
Sorted!

illicitonion:

PR #352, opened 2026-03-11
London | 26-SDC-Mar | Beko | Sprint 1 | Individual shell tools exercises

Abubakar-Meigag:
Self checklist

## Changelist

complete individual shell tools exercises

## Questions

no questions, thanks!

### DIFF
--- a/individual-shell-tools/awk/script-01.sh
+++ b/individual-shell-tools/awk/script-01.sh
@@ -2,5 +2,7 @@
 
 set -euo pipefail
 
+awk '{print NF}' scores-table.txt
+
 # TODO: Write a command to output just the names of each player in `scores-table.txt`.
 # Your output should contain 6 lines, each with just one word on it.
--- a/individual-shell-tools/awk/script-01.sh
+++ b/individual-shell-tools/awk/script-01.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-awk '{print NF}' scores-table.txt
+awk '{print $1}' scores-table.txt
 
 # TODO: Write a command to output just the names of each player in `scores-table.txt`.
 # Your output should contain 6 lines, each with just one word on it.
--- a/individual-shell-tools/awk/script-04.sh
+++ b/individual-shell-tools/awk/script-04.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+awk '/London/ {print $1, $4}' scores-table.txt
+
 # TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
 # Your output should contain 3 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4".
--- a/individual-shell-tools/awk/script-04.sh
+++ b/individual-shell-tools/awk/script-04.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-awk '/London/ {print $1, $4}' scores-table.txt
+awk '$2 == "London" {print $1, $NF}' scores-table.txt
 
 # TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
 # Your output should contain 3 lines, each with one word and one number on it.
--- a/individual-shell-tools/awk/script-07-stretch.sh
+++ b/individual-shell-tools/awk/script-07-stretch.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+awk '{for (i=3; i<=NF; i++) sum[$1] += $i} END {for (name in sum) print name, sum[name]}' scores-table.txt
+
 # NOTE: This is a stretch exercise - it is optional.
 
 # TODO: Write a command to output just the names of each player along with the total of adding all of that player's scores.
--- a/individual-shell-tools/awk/script-07-stretch.sh
+++ b/individual-shell-tools/awk/script-07-stretch.sh
@@ -2,7 +2,12 @@
 
 set -euo pipefail
 
-awk '{for (i=3; i<=NF; i++) sum[$1] += $i} END {for (name in sum) print name, sum[name]}' scores-table.txt
+awk '{
+  if (!seen[$1]) { order[++count] = $1; seen[$1] = 1 }
+  for (i=3; i<=NF; i++) sum[$1] += $i
+} END {
+  for (i=1; i<=count; i++) print order[i], sum[order[i]]
+}' scores-table.txt
 
 # NOTE: This is a stretch exercise - it is optional.
 
--- a/individual-shell-tools/grep/script-04.sh
+++ b/individual-shell-tools/grep/script-04.sh
@@ -2,5 +2,8 @@
 
 set -euo pipefail
 
+
+grep -v "Hello" dialogue.txt
+
 # TODO: Write a command to output every line in dialogue.txt that does not contain the word "Hello" (regardless of case).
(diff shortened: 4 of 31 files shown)


individual-shell-tools/awk/script-01.sh
```diff
@@ -2,5 +2,7 @@
 
 set -euo pipefail
 
+awk '{print NF}' scores-table.txt
```
SlideGauge: What does NF mean in this context and what will this line output?

individual-shell-tools/awk/script-04.sh
```diff
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+awk '/London/ {print $1, $4}' scores-table.txt
```
SlideGauge: What happens when we meet lines with different number of scores?

individual-shell-tools/grep/script-04.sh
```diff
@@ -2,5 +2,8 @@
 
 set -euo pipefail
 
+
+grep -v "Hello" dialogue.txt
```
SlideGauge: Will it find lines like "hello" (lowercase)?

individual-shell-tools/ls/script-01.sh
```diff
@@ -13,3 +13,5 @@ fi
 
 # TODO: Write a command to list the files and folders in this directory.
 # The output should be a list of names including child-directory, script-01.sh, script-02.sh, and more.
+
+ls ../ls
```
SlideGauge: README file says
"You should write all of your scripts assuming they're running inside the directory they're saved in."
Whilst this line works, it does redundant operation with the path, could you simplify  it?

individual-shell-tools/ls/script-02.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command which lists all of the files in the directory named child-directory.
 # The output should be a list of names: helper-1.txt, helper-2.txt, helper-3.txt.
+
+ls ../ls/child-directory
```
SlideGauge: README file says
"You should write all of your scripts assuming they're running inside the directory they're saved in."
Whilst this line works, it does redundant operation with the path, could you simplify  it?

individual-shell-tools/ls/script-03.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # TODO: Write a command which _recursively_ lists all of the files and folders in this directory _and_ all of the files inside those folders.
 # The output should be a list of names including: child-directory, script-01.sh, helper-1.txt (and more).
 # The formatting of the output doesn't matter.
+
+ls ../ls ../ls/child-directory
```
SlideGauge: README file says
"You should write all of your scripts assuming they're running inside the directory they're saved in."
Whilst this line works, it does redundant operation with the path, could you simplify  it?

individual-shell-tools/ls/script-04.sh
```diff
@@ -12,12 +12,12 @@ sleep 1
 touch "${script_dir}/child-directory/helper-3.txt"
 
 echo "First exercise (sorted newest to oldest):"
-
+ls -t ../ls/child-directory
```
SlideGauge: README file says
"You should write all of your scripts assuming they're running inside the directory they're saved in."
Whilst this line works, it does redundant operation with the path, could you simplify  it?

individual-shell-tools/ls/script-03.sh
SlideGauge: Also, will it output directories contents recursively?

individual-shell-tools/grep/script-06.sh
```diff
@@ -2,5 +2,7 @@
 
 set -euo pipefail
 
+grep -l "Doctor:" *.txt
```
SlideGauge: Will it output only lines of the Doctor or can it potentially output lines which contain "Doctor:" somewhere in the middle?

individual-shell-tools/awk/script-04.sh
SlideGauge: What will happen if some of the player's have name "London" but the city is not London?

individual-shell-tools/awk/script-07-stretch.sh
```diff
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+awk '{for (i=3; i<=NF; i++) sum[$1] += $i} END {for (name in sum) print name, sum[name]}' scores-table.txt
```
SlideGauge: Does `for (name in sum)` guarantee the order of traversal?

SlideGauge (commented):
There a couple of files which do not have empty line in the end of the file (like wc/script-02.sh). Could you add them please

SlideGauge:
Wrote my notes, could you fulfil them please?

individual-shell-tools/awk/script-01.sh
Abubakar-Meigag: NF counts the number of fields in each row, so it was outputting the word count per line rather than the names

I've updated the command to use $1 instead, which prints the value of the first field.

individual-shell-tools/awk/script-04.sh
Abubakar-Meigag: $4 is hardcoded so it won't work for players with different number of scores
i should use $NF instead which always points to the last field.

individual-shell-tools/awk/script-07-stretch.sh
Abubakar-Meigag: actually no, for (name in sum) doesn't guarantee order
because awk arrays are associative so traversal is arbitrary

i've fixed it by tracking the names in a separate ordered array

individual-shell-tools/grep/script-04.sh
Abubakar-Meigag: No, it won't catch lowercase 'hello' because grep is case-sensitive by default.

i've added the -i flag to make it case-insensitive

individual-shell-tools/grep/script-06.sh
Abubakar-Meigag: It can match 'Doctor:' anywhere in the line

so I've added ^ to anchor it to the start.

individual-shell-tools/ls/script-01.sh
Abubakar-Meigag: Since the script runs in its own directory

i can just use ls with no path.

updated

individual-shell-tools/ls/script-02.sh
Abubakar-Meigag: updated

individual-shell-tools/ls/script-03.sh
Abubakar-Meigag: I've simplified it to ls -R

which recursively lists everything from the current directory

Abubakar-Meigag: updated

individual-shell-tools/ls/script-04.sh
Abubakar-Meigag: updated

Abubakar-Meigag:
Hi @SlideGauge thanks for the review, I've completed all the changes for the code review and feedback

Thanks!

SlideGauge:
Review complete, good job!

illicitonion:

PR #360, opened 2026-03-12
Glasgow | 26- SDC-Mar | Taras Mykytiuk | Sprint 1 | Number systems

TarasMykytiuk:
## Learners, PR Template

Self checklist

## Changelist

Number systems exercises are done.

### DIFF
--- a/number-systems/README.md
+++ b/number-systems/README.md
@@ -5,61 +5,61 @@ Do not convert any binary numbers to decimal when solving a question unless the 
 The goal of these exercises is for you to gain an intuition for binary numbers. Using tools to solve the problems defeats the point.
 
 Convert the decimal number 14 to binary.
-Answer:
+Answer: 1110
 
 Convert the binary number 101101 to decimal:
-Answer:
+Answer: 45
 
 Which is larger: 1000 or 0111?
-Answer:
+Answer: 1000
 
 Which is larger: 00100 or 01011?
-Answer:
+Answer: 01011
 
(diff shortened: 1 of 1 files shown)


cjyuan (commented):
Can you do a self-check against the answers on https://github.com/CodeYourFuture/Module-Tools/blob/solutions/number-systems/README.md?plain=1

Should you have any question regarding number systems, feel free to raise them in the comment of this PR.

TarasMykytiuk:
> Can you do a self-check against the answers on https://github.com/CodeYourFuture/Module-Tools/blob/solutions/number-systems/README.md?plain=1
> 
> Should you have any question regarding number systems, feel free to raise them in the comment of this PR.

Hi @cjyuan . Thanks for review. I have fixed conversion from hex to decimal.
I compared my other answers to self-check.
Is that answer enough?:

"How can you test if a binary number is a power of two (e.g. 1, 2, 4, 8, 16, ...)?
Answer: Test that it has exactly one 1."

cjyuan:
Change looks good. 

> "How can you test if a binary number is a power of two (e.g. 1, 2, 4, 8, 16, ...)?
Answer: Test that it has exactly one 1."

Aren't you curious how to implement such a check in JS or in Python?

cjyuan:
The approach involving `log()` function may not be 100% accurate because of floating point arithmetic, and is relatively slower than an approach that involves only one bit-wise and one arithmetic operations.

illicitonion:

PR #366, opened 2026-03-13
Sheffield | 2026-MAR-SDC | Jak Rhodes-Smith | Sprint 3 | Implement shell tools

jakr-s:
## Learners, PR Template

Self checklist

## Changelist

Complete all NodeJS tool implementations

Task ID: CYF-1150

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -4,14 +4,27 @@ import * as fs from "node:fs/promises";
 program
   .name("cat")
   .description("Reads file and writes it to the standard output")
-  .argument("<path>", "The file path to process");
+  .argument("<path>", "The file path to process")
+  .option("-n", "Number the output lines, starting at 1.")
+  .option("-b", "Number only non-blank output lines, starting at 1.");
 
 program.parse();
 
 try {
   const [filePath] = program.args;
-  const contents = await fs.readFile(filePath, { encoding: "utf8" });
-  console.log(contents);
+  const options = program.opts();
+
+  const file = await fs.open(filePath);
+
+  let lineNum = 1;
+  for await (const line of file.readLines()) {
+    const isBlank = line.trim() === "";
+    const shouldNumber = options.n || (options.b && !isBlank);
+
+    console.log(shouldNumber ? `${lineNum}  ${line}` : line);
+
+    if (shouldNumber) lineNum++;
+  }
 } catch (err) {
   console.error(err.message);
 }
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,0 +1,17 @@
+import { program } from "commander";
+import * as fs from "node:fs/promises";
+
+program
+  .name("cat")
+  .description("Reads file and writes it to the standard output")
+  .argument("<path>", "The file path to process");
+
+program.parse();
+
+try {
+  const [filePath] = program.args;
+  const contents = await fs.readFile(filePath, { encoding: "utf8" });
+  console.log(contents);
+} catch (err) {
+  console.error(err.message);
+}
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -3,27 +3,35 @@ import * as fs from "node:fs/promises";
 
 program
   .name("cat")
-  .description("Reads file and writes it to the standard output")
-  .argument("<path>", "The file path to process")
+  .description("Reads file(s) and writes them to the standard output")
+  .argument("<paths...>", "The file path(s) to process")
   .option("-n", "Number the output lines, starting at 1.")
   .option("-b", "Number only non-blank output lines, starting at 1.");
 
 program.parse();
 
 try {
-  const [filePath] = program.args;
+  const filePaths = program.args;
+
   const options = program.opts();
 
-  const file = await fs.open(filePath);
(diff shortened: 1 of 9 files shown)


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,38 @@
...
+    const file = await fs.open(filePath);
+
+    let lineNum = 1;
+
+    try {
+      for await (const line of file.readLines()) {
+        const isBlank = line.trim() === "";
+        const shouldNumber = options.n || (options.b && !isBlank);
+
+        console.log(shouldNumber ? `${lineNum}  ${line}` : line);
```
Poonam-raj, line 28: To match the `cat` output accurately you would need to add the correct padding instead of just spaces. Here it's meant to be padding of 6 so the output is right-aligned.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,38 @@
...
+
+        if (shouldNumber) lineNum++;
+      }
+    } finally {
+      await file.close();
```
Poonam-raj, line 33: Nice use of finally to close the file after reading

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,67 @@
...
+    "The file path to process (defaults to current directory)",
+  )
+  .option("-a", "Include directory entries whose names begin with a dot ('.').")
+  .option("-1", "Force output to be one entry per line.");
+
+program.parse();
+
+try {
+  let filePaths = program.args;
+  if (!filePaths || filePaths.length === 0) {
```
Poonam-raj, line 18: I would argue `!filePaths` is a bad path you weren't asked to account for

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  const filterHidden = (files) => files.filter((file) => !file.startsWith("."));
+
+  const getVisibleEntries = (files) =>
+    includeHidden ? files : filterHidden(files);
+
+  const formatEntries = (files) => {
+    if (files.length === 0) return;
+    console.log(files.join(onePerLine ? "\n" : "\t"));
```
Poonam-raj, line 43: I agree the use of `\t` is an easier fix to calculating the column padding needed, but maybe less accurate to the real `ls` output columns.

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,67 @@
...
+    const stats = await fs.stat(filePath);
+    if (stats.isFile()) result.files.push(filePath);
+    if (stats.isDirectory()) {
+      result.dirs[filePath] = await fs.readdir(filePath);
+    }
+  }
+
+  const filterHidden = (files) => files.filter((file) => !file.startsWith("."));
+
+  const getVisibleEntries = (files) =>
```
Poonam-raj, line 38: I personally dislike having a function like `getVisibleEntries` be dependent on a variable outside of this function. I'd prefer you to pass in `includeHidden`

It breaks the idea of "purity" (a function's output depends solely on arguments passed in or internal logic, no external dependencies - which `includeHidden` is a hidden dependency). It makes it hard to test (we need dependencies more loosely coupled), and harder to reuse

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,67 @@
...
+    console.log(files.join(onePerLine ? "\n" : "\t"));
+  };
+
+  result.files = getVisibleEntries(result.files);
+
+  if (filePaths.length === 1) {
+    let entries = [...result.files];
+
+    for (const [dir, contents] of Object.entries(result.dirs)) {
+      const filtered = getVisibleEntries(contents);
```
Poonam-raj, line 52: (based on previous comment) we can see here how `getVisibleEntries` leaves a lot of questions - how is it determining what is visible or not? Is this a hard coded process, or is it a changeable output?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,67 @@
...
+      entries = entries.concat(filtered);
+    }
+    formatEntries(entries);
+  } else {
+    formatEntries(result.files);
+
+    for (const [dir, contents] of Object.entries(result.dirs)) {
+      console.log("\n" + dir + ":");
```
Poonam-raj, line 60: Small bug here - you're baking into your output a newline on the first line of prints - not an expected behaviour, the first line should be at the top with no newline

How can you separate out the concern so the newline gaps are **between directory sections only** and the top section doesn't have a newline before it?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,65 @@
...
+    results["total"] = total;
+  }
+
+  const options = program.opts();
+  const noOptionsProvided = !Object.keys(options).length;
+  const selectedOptionKeys = [...Object.keys(options)];
+
+  noOptionsProvided
+    ? console.table(results)
+    : console.table(results, selectedOptionKeys);
```
Poonam-raj, line 62: Only feedback here is that console.table doesn't achieve the same output that the command would. Padding would more achieve the right look here

Poonam-raj (commented):
A few areas for improvement I would say the things I'd want fixed before approving would be:
- the newlines in the ls command (between directories only)
- The purity of the `getVisibleEntries` function needs amending 
- Adding padding to the output of the cat command instead of using spaces

PR #367, opened 2026-03-13
London | 26-SDC-Mar | Jamal Laqdiem | Sprint 2 | shell pipelines exercises

jamallaqdiem:
Self checklist

- [x ] My changes meet the requirements of the task

## Changelist
Completed the exercises shell pipelines.


shell-pipelines/ls-grep/script-03.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+ls -1 sample-files | grep '^[A-Z][a-z]*$'
```
SlideGauge: The regex '^[A-Z][a-z]*$' captures "starts with uppercase, then only lowercase letters." But the task says "doesn't contain any other upper case letters" — it says nothing about restricting digits, dots, hyphens, etc. A filename like Foo.txt or File1 would NOT match your regex even though it satisfies the task requirement.

shell-pipelines/ls-grep/script-04.sh
```diff
@@ -4,3 +4,6 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
+#!/bin/bash
+
+ls -1 sample-files | grep '^[A-Z][a-z]*$' | wc -l
```
SlideGauge: Same regex concern as script-03

SlideGauge (commented):
Could you adjust the settings of your editor? All the files are missing the trailing end of line.

SlideGauge:
Good job, could you do a couple of fixes I pointed out?

shell-pipelines/ls-grep/script-03.sh
jamallaqdiem: Thanks for reviewing

shell-pipelines/ls-grep/script-04.sh
jamallaqdiem: Thnaks for reviewing, I addressed the issues.

jamallaqdiem:
I changed the setting to add a trainling end line at save.

SlideGauge:
Nice, review completed

illicitonion:

PR #368, opened 2026-03-13
London | 26-SDC-Mar | Craig D'Silva | Sprint 1 | Individual shell tools

craig-dsilva:
## Learners, PR Template

Self checklist

## Changelist

This is my coursework for Individual shell tools exercise.

### DIFF
--- a/individual-shell-tools/awk/script-04.sh
+++ b/individual-shell-tools/awk/script-04.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
 # Your output should contain 3 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4".
+
+awk '{if ($2 == "London")  print $1, $NF}' scores-table.txt
--- a/individual-shell-tools/awk/script-06-stretch.sh
+++ b/individual-shell-tools/awk/script-06-stretch.sh
@@ -6,3 +6,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the total of adding together all players' first scores.
 # Your output should be exactly the number 54.
+
+awk '{ score += $3 } END { print score }' scores-table.txt
--- a/individual-shell-tools/cat/script-03.sh
+++ b/individual-shell-tools/cat/script-03.sh
@@ -9,3 +9,5 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+
+cat ../helper-files/helper-3.txt -n
--- a/individual-shell-tools/cat/script-03.sh
+++ b/individual-shell-tools/cat/script-03.sh
@@ -10,4 +10,4 @@ set -euo pipefail
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
 
-cat ../helper-files/helper-3.txt -n
+cat -n ../helper-files/helper-3.txt
--- a/individual-shell-tools/cat/script-04-stretch.sh
+++ b/individual-shell-tools/cat/script-04-stretch.sh
@@ -13,3 +13,5 @@ set -euo pipefail
 # 3 It looked delicious.
 # 4 I was tempted to take a bite of it.
 # 5 But this seemed like a bad idea...
+
+cat ../helper-files/*.txt -n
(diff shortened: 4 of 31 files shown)


individual-shell-tools/grep/script-06.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the name of every `.txt` file in this directory which contains a line of dialogue said by the Doctor.
 # The output should contain two filenames.
+
+grep ^Doctor *.txt
```
SlideGauge: 1. What will happen if this pattern faces a line looking like this:
DoctorImpostor: I'm real!
2. The task asks to output filenames only. What will happen our case, though?

individual-shell-tools/grep/script-07.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output, for each `.txt` file in this directory, how many lines of dialogue the Doctor has.
 # The output should show that dialogue.txt contains 6 lines, dialogue-2.txt contains 2, and dialogue-3.txt contains 0.
+
+grep -c ^Doctor *.txt
```
SlideGauge, line 8: Same here about the line like:
DoctorImpostor: lalalalala

individual-shell-tools/grep/script-01.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt said by the Doctor.
 # The output should contain 6 lines.
+
+grep ^Doctor dialogue.txt
```
SlideGauge, line 8: What will happen with the line like:
DoctorHouse: la la la

individual-shell-tools/awk/script-06-stretch.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the total of adding together all players' first scores.
 # Your output should be exactly the number 54.
+
+awk '{ score += $3 } END { print score }' scores-table.txt
```
SlideGauge, line 10: good name for the variable name, self-descriptive, like

individual-shell-tools/awk/script-04.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player in London along with the score from their last attempt.
 # Your output should contain 3 lines, each with one word and one number on it.
 # The first line should be "Ahmed 4".
+
+awk '{if ($2 == "London")  print $1, $NF}' scores-table.txt
```
SlideGauge, line 9: Correct use of NF, like

individual-shell-tools/ls/script-04.sh
```diff
@@ -16,8 +16,12 @@ echo "First exercise (sorted newest to oldest):"
 # TODO: Write a command which lists the files in the child-directory directory, one per line, sorted so that the most recently modified file is first.
 # The output should be a list of names in this order, one per line: helper-3.txt, helper-1.txt, helper-2.txt.
 
+ls child-directory -t
```
SlideGauge, line 19: Whilst it indeed produces what is asked, do we have a way explicitly tell ls to output list of names one per line?

individual-shell-tools/cat/script-03.sh
```diff
@@ -9,3 +9,5 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+
+cat ../helper-files/helper-3.txt -n
```
SlideGauge: it won't affect the result in this case, but just for clarity, it is better to pass options (i.e. -n) before filenames.
Typically the format of usage is
command [options] [filenames]

individual-shell-tools/cat/script-04-stretch.sh
```diff
@@ -13,3 +13,5 @@ set -euo pipefail
 # 3 It looked delicious.
 # 4 I was tempted to take a bite of it.
 # 5 But this seemed like a bad idea...
+
+cat ../helper-files/*.txt -n
```
SlideGauge: it won't affect the result in this case, but just for clarity, it is better to pass options (i.e. -n) before filenames.

SlideGauge:
Overall good job, could you address my notes please?

individual-shell-tools/ls/script-04.sh
craig-dsilva: Yes

individual-shell-tools/grep/script-06.sh
craig-dsilva: This will print the matching patterns also. I have now added the flag to suppress the pattern.

SlideGauge:
Review complete

PR #376, opened 2026-03-14
London | 26-SDC-Mar | Ebrahim Beiati-Asl | Sprint 3 | implements-shell-tool

ebrahimbeiati:
## Self checklist

## Changelist
Exercise for implement-shell-tools


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,76 @@
...
+        } else if(arg === '-b') {
+            options.numberNonEmpty = true;
+        } else {
+            filePatterns.push(arg);
+        }
+    });
+    // -b takes precedence over -n
+    if(options.numberNonEmpty) {
+        options.numberAll = false;
+    }
```
illicitonion: Rather than having two different options which are mutually exclusive, I would probably model this as one option with three possible values: `options.numberMode = "off" | "all" | "non-empty"`. That way you don't need to re-set `numberAll` if actually `numberNonEmpty` - you just have one variable you set correctly up-front and which you can check.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,76 @@
...
+
+    if(filePatterns.length === 0) {
+        console.log("cat: missing file operand");
+        process.exit(1);
+    }
+
+    const files = filePatterns;
```
illicitonion: Why did you call this `filePatterns` just to rename it to `files` later? Why not just name it `files` or `fileNames` from the start?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,76 @@
...
+
+    files.forEach((file) => {
+        const resolvedPath = path.resolve(process.cwd(), file);
```
illicitonion: What advantage does `path.resolve(process.cwd(), file)` have over just using `file` directly?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+        const entries = fs.readdirSync(dirPath, { withFileTypes: true });
+        const filtered = entries.filter((entry) => showAll || !entry.name.startsWith('.'));
+
+        if(onePerLine) {
+            filtered.forEach(entry => console.log(entry.name));
+        } else {
+            const names = filtered.map(entry => entry.name);
+            console.log(names.join('  '));
+        }
+        } catch (error) {
```
illicitonion: This indentation isn't correct - please format all of your code before submitting for review. I'd recommend setting up your IDE to auto-format your code on save to avoid needing to remember to do this.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+//print counts in the format of wc according to options
+function printCounts(filePath, counts, options) {
+    const parts = [];
+    if(options.line) parts.push(counts.lines);
+    if(options.word) parts.push(counts.words);
+    if(options.byte) parts.push(counts.bytes);
+//if no specific count options are provided, print all counts
+    if(!options.line && !options.word && !options.byte) {
```
illicitonion: It _may_ be simpler to do this check when setting up options, rather than when consuming it - i.e. in your options parsing to see "if no options were set, set all to true" - that way the rest of your code doesn't need to think about this special case, they can just handle `wc foo` and `wc -l -w -c foo` exactly the same.

(This about this as an _abstraction boundary_ - the point of parsing command line flags into an options object is so that the rest of your code doesn't need to worry about the specifics of flags, this is just one more example of a place you can avoid your code needing to know the specifics of flags)

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+function main() {
+    const args = process.argv.slice(2);
+    // Check for options
+    const showAll = args.includes('-a');
+    const onePerLine = args.includes('-1');
+    //remove options from args 
+    const directories = args.filter(arg => arg !== '-a' && arg !== '-1');
+
+    // If no directory is specified, list the current directory
+    if(directories.length === 0) {
```
illicitonion: Can you think how to fold the "no directories" case into the main loop so you don't have to have two separate calls to `listDirectory` in this function?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,76 @@
...
+                //-n option: number all lines
+                process.stdout.write(
+                    `${String(globalLineCounter).padStart(6)}\t${line}\n`
+                );
+                globalLineCounter++;
+            } else {
+                //default: just print the line
+                process.stdout.write(line + '\n');
+            }
+    });
```
illicitonion: The three branches here look quite similar and repetitive. In general, if you have multiple similar branches, it's more clear to extract the differences into variables, and then run the same code, i.e. so you'd only have one call to `process.stdout.write` which looks more like ``process.stdout.write(`${prefix}${line}\n`)`` where `prefix` may be set differently based on options (including potentially an empty string). 

illicitonion (commented):
This generally looks good, but I left a few comments to consider. Well done!

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,60 @@
...
+function printFile(filePath, options) {
+  try {
+    const content = fs.readFileSync(filePath, "utf-8");
+    const lines = content.split("\n");
+
+    lines.forEach((line) => {
+      let prefix = "";
+
+      if (options.numberMode === "non-empty") {
+        //-b option: number non-empty lines
```
illicitonion: This comment feels redundant - reading the if condition tells me the same information as the comment does

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,60 @@
...
+        if (line.trim() !== "") {
+          prefix = `${String(globalLineCounter).padStart(6)}\t`;
+          globalLineCounter++;
+        }
+      } else if (options.numberMode === "all") {
+        prefix = `${String(globalLineCounter).padStart(6)}\t`;
+        globalLineCounter++;
```
illicitonion: There's still some repetition here in setting up the prefix. How about something like:

```js
if (options.numberMode === "all" || (options.numberMode === "non-empty" && line.trim() !== "") {
```

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+    const entries = fs.readdirSync(dirPath, { withFileTypes: true });
+    const filtered = entries.filter(
+      (entry) => showAll || !entry.name.startsWith("."),
+    );
+
+    if (onePerLine) {
+      filtered.forEach((entry) => console.log(entry.name));
+    } else {
+      const names = filtered.map((entry) => entry.name);
+      console.log(names.join("  "));
```
illicitonion, line 14: In `cat` you used `process.stdout.write` and here you're using `console.log` - what's the difference? When would you choose to use one or the other?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  } catch (error) {
+    console.error(`Error reading directory ${dirPath}: ${error.message}`);
```
illicitonion, line 17: What exit code should our process terminate with if it hit an error? What does your code do?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+  const showAll = args.includes("-a");
+  const onePerLine = args.includes("-1");
+  //remove options from args
+  let directories = args.filter((arg) => arg !== "-a" && arg !== "-1");
+
+  // Default to current directory if no directories are specified
+  if (directories.length === 0) {
+    directories = [process.cwd()];
+  }
+  //If a directory is specified, list that directory
```
illicitonion: This comment seems a bit confusing? I'd remove it

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+    try {
+      const stats = fs.statSync(arg);
+
+      if (stats.isDirectory()) {
+        //Print header if multiple directories are listed
+        if (directories.length > 1) console.log(`${arg}:`);
+        listDirectory(arg, showAll, onePerLine);
+        //add a blank line between directory listings if there are multiple directories
+        if (directories.length > 1 && index < directories.length - 1)
+          console.log("");
```
illicitonion: I'd recommend always using `{}`s around if conditions, or perhaps sometimes allowing single-line statements (like your above ``if (directories.length > 1) console.log(`${arg}:`);``) - see https://www.blackduck.com/blog/understanding-apple-goto-fail-vulnerability-2.html for an example of why

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,73 @@
...
+      const counts = countFileContent(content);
+
+      // Sum counts for total if multiple files
+      totalCounts.lines += counts.lines;
+      totalCounts.words += counts.words;
+      totalCounts.bytes += counts.bytes;
+
+      printCounts(file, counts, options);
+    } catch (error) {
+      console.error(`Error reading file ${file}: ${error.message}`);
```
illicitonion, line 65: Same question about exit codes

illicitonion (commented):
Looking really good, great improvements! A few more small things to think about :)

implement-shell-tools/ls/ls.js
ebrahimbeiati: We use `process.stdout.write` when we need full control over spacing and newlines, byte-exact output matters, and to avoid automatic newlines, like in cat.
We use `console.log` when we just want simple, line‑by‑line output with automatic newlines, like in ls.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,58 @@
...
+
+        if (shouldNumber) {
+          prefix = `${String(globalLineCounter).padStart(6)}\t`;
+          globalLineCounter++;
+        }
+
+      process.stdout.write(`${prefix}${line}\n`);
+    });
+  } catch (error) {
+    console.error(`cat: ${filePath}: ${error.message}`);
```
illicitonion, line 30: This will still exit 0 if e.g. a file didn't exist.

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,57 @@
...
+  // Default to current directory if no directories are specified
+  if (directories.length === 0) {
+    directories = [process.cwd()];
+  }
+  directories.forEach((arg, index) => {
+    try {
+      const stats = fs.statSync(arg);
+
+      if (stats.isDirectory()) {
+        //Print header if multiple directories are listed
```
illicitonion: Is this comment useful compared to just reading the code?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,57 @@
...
+}
+function main() {
+  const args = process.argv.slice(2);
+  // Check for options
+  const showAll = args.includes("-a");
+  const onePerLine = args.includes("-1");
+  //remove options from args
+  let directories = args.filter((arg) => arg !== "-a" && arg !== "-1");
+
+  // Default to current directory if no directories are specified
```
illicitonion: Is this comment useful compared to just reading the code?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+function main() {
```
illicitonion, line 35: This tool behaves slightly differently than built-in `cat` for the examples given in https://github.com/CodeYourFuture/Module-Tools/blob/main/implement-shell-tools/cat/README.md

illicitonion: Your `wc` does too for the same reason.

illicitonion (commented):
Looks good! Just a few last things :)

implement-shell-tools/cat/cat.js
ebrahimbeiati: Thank you so much for that. I'll write an integration test to find the issue.

illicitonion (commented):
Looks great, well done!

FWIW the integration test looks really useful, and is probably worth re-adding, just inside the `implement-shell-tools` directory rather than the root of the repo :)

illicitonion:

PR #379, opened 2026-03-15
Birmingham | 2026-MAR-SDC | Joy Opachavalit | Sprint 2 | jq

enjoy15:
## Learners, PR Template

Self checklist

## Changelist

Completed all exercises

### DIFF
--- a/jq/script-11.sh
+++ b/jq/script-11.sh
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the total of adding together all scores from all games from all players.
 # Your output should be exactly the number 164.
+jq -r 'map(.scores) | add | add' scores.json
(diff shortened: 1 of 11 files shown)


jq/script-11.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the scores.json file.
 # TODO: Write a command to output the total of adding together all scores from all games from all players.
 # Your output should be exactly the number 164.
+jq -r 'map(.scores) | add | add' scores.json
```
cjyuan, line 8: Can you explain how the two `add` works in the query? That is, why "add" twice?

cjyuan (commented):
Solution looks good. I just have a question.

jq/script-11.sh
enjoy15: Data shape: scores.json is an array of player objects where each scores value is an array (e.g. [[1,10,4],[22,9,6],...]).
map(.scores) produces an array of arrays: [[1,10,4],[22,9,6],...].
First add: when add is given an array of arrays it concatenates them into one array: [1,10,4,22,9,6,...].
Second add: when add is given an array of numbers it returns their sum (i.e. 164).
So map(.scores) | add | add is: collect each player's scores then join into one big list then sum the numbers.

cjyuan: All good. Well done.

illicitonion:

PR #382, opened 2026-03-15
Glasgow | 26- SDC-Mar | Taras Mykytiuk | Sprint 1 | Individual shell tools

TarasMykytiuk:
## Learners, PR Template

Self checklist

## Changelist

Individual shell tools exercises are done.

### DIFF
--- a/individual-shell-tools/awk/script-07-stretch.sh
+++ b/individual-shell-tools/awk/script-07-stretch.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 # NOTE: This is a stretch exercise - it is optional.
-
+awk '{sum=o; for(i=3; i<=NF; i++) sum+=$i; print $1, sum}' scores-table.txt
 # TODO: Write a command to output just the names of each player along with the total of adding all of that player's scores.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15". The second line should be "Basia 37"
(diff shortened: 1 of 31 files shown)


individual-shell-tools/ls/script-03.sh
```diff
@@ -3,5 +3,6 @@
 set -euo pipefail
 
 # TODO: Write a command which _recursively_ lists all of the files and folders in this directory _and_ all of the files inside those folders.
+ls -r
```
SlideGauge: What does the small -r stand for? Will it lead to expected behaviour here?

individual-shell-tools/awk/script-07-stretch.sh
```diff
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 # NOTE: This is a stretch exercise - it is optional.
-
+awk '{sum=o; for(i=3; i<=NF; i++) sum+=$i; print $1, sum}' scores-table.txt
```
SlideGauge, line 6: Could you doublecheck the initial value of sum we initialise it with?

individual-shell-tools/grep/script-01.sh
```diff
@@ -3,4 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt said by the Doctor.
+grep  "^Doctor:" dialogue.txt
```
SlideGauge, line 6: Good use of ^Doctor: to anchor the match to the start of the line — this correctly avoids matching lines where    "Doctor" appears in the middle of dialogue.

SlideGauge:
Good work, I just have two tiny notes, could you fix them?

individual-shell-tools/ls/script-03.sh
TarasMykytiuk: Thanks. It is easy to overlook.

illicitonion:

PR #384, opened 2026-03-16
London| 2026-MAR-SDC | Imran Mohamed | Sprint 2 | Shell pipelines

i786m:
Self checklist

## Changelist

Added commands for shell pipelines exercises

### DIFF
--- a/shell-pipelines/sort-uniq-head-tail/script-06.sh
+++ b/shell-pipelines/sort-uniq-head-tail/script-06.sh
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+sort events.txt | uniq -c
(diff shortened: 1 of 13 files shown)


shell-pipelines/sort-uniq-head-tail/script-06.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+sort events.txt | uniq -c
```
SlideGauge, line 8: This command groups by the full line, so the output is per-person counts: 2 Entry German, 1 Entry Mariana, 2 Entry Sally, etc. The task says "it should be clear from your script's output that there have been 5 Entry events and 4 Exit events" - a reader has to sum the numbers mentally to get there. You need to count by event type only.

shell-pipelines/sort-uniq-head-tail/script-07.sh
```diff
@@ -6,3 +6,4 @@ set -euo pipefail
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
 # The word "Event" should not appear in your script's output.
+sort events-with-timestamps.txt | uniq -c | grep -v "Event"
```
SlideGauge, line 9: Because every timestamp+person combination in events-with-timestamps.txt is unique, sort | uniq -c just   outputs every line with a count of 1 - it doesn't aggregate Entry vs Exit.

SlideGauge (commented):
Could you check trailing new line presence in the files? Some files does not have it, github highlights it with 

Image

SlideGauge:
Could you address my comments please?

PR #385, opened 2026-03-16
NW | 2026-mar-sdc | Zabihollah Namazi | Sprint 2 | shell pipelines

ZabihollahNamazi:
Self checklist

## Changelist

shell pipeline exercises are done and please if you could check them , thanks


shell-pipelines/ls-grep/script-03.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+ls sample-files | grep '^[A-Z][a-z]*$'    # [a-z]* = zero or more lowercase after that , $ = end of this line grep
```
SlideGauge: Will this regex pass such a filename as "File2.txt"?

shell-pipelines/ls-grep/script-04.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
+ls sample-files | grep -c '^[A-Z][a-z]*$'    # -c to count lines
```
SlideGauge: Same question as script-03.sh

shell-pipelines/sort-uniq-head-tail/script-06.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+awk '{print $1}' events.txt | sort | uniq -c
```
SlideGauge: Good catch of Entry/Exit events via $1

SlideGauge (commented):
Could you adjust settings of your editor? All the scripts does not have trailing new line

SlideGauge:
Good job, just a couple of notes from me to fix please

shell-pipelines/ls-grep/script-03.sh
ZabihollahNamazi: no it doesnt and i fix it which is 
[^A-Z]*$     zero or more letter after that first uppercas which is not upper case
thanks

shell-pipelines/ls-grep/script-04.sh
ZabihollahNamazi: no it doesnt and i fix it which is 
[^A-Z]*$     zero or more letter after that first uppercas which is not upper case
thanks

illicitonion:

PR #389, opened 2026-03-19
NW | 26-SDC-Mar | TzeMing Ho | Sprint 2 | jq exercises

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

I have completed the interesting jq exercises.

### DIFF
--- a/jq/script-09.sh
+++ b/jq/script-09.sh
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total scores from all of their games added together.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15" with no quotes.
+
+jq -r '.[] | [.name, (.scores | reduce .[] as $score (0; . + $score))] | join(" ")' ./scores.json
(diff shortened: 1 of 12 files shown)


jq/script-09.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total scores from all of their games added together.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15" with no quotes.
+
+jq -r '.[] | [.name, (.scores | reduce .[] as $score (0; . + $score))] | join(" ")' ./scores.json
```
cjyuan: Why not use `add`?

cjyuan (commented):
Your solution looks good.

You may want to check out these alternatives (especially those that use string interpolation):
https://github.com/CodeYourFuture/Module-Tools/tree/solutions/jq

I only have one question.

jq/script-09.sh
TzeMingHo: I was trying to explore the method that I have known from JavaScript, and didn't know I could have used add. Indeed, using add is much cleaner.

cjyuan: Do keep on trying different approaches. That's how we learn. (Just need to submit the best one in a PR). :)

All good.

illicitonion:

PR #391, opened 2026-03-19
NW | 26-SDC-Mar | TzeMing Ho | Sprint 2 | shell pipelines exercises

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

I have completed the exercises on shell pipelines.

### DIFF
--- a/shell-pipelines/ls-grep/script-03.sh
+++ b/shell-pipelines/ls-grep/script-03.sh
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+
+ls ./sample-files | grep '^[A-Z][a-z]*$'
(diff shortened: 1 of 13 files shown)


shell-pipelines/ls-grep/script-03.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+
+ls ./sample-files | grep '^[A-Z][a-z]*$'
```
SlideGauge: Will it pass such filenames as File2 or File3.txt?

SlideGauge (commented):
Could you add trailing new lines to each of the script? Try to adjust your editor's settings.

SlideGauge:
Good job, just a couple of notes from my side, fix them please.

shell-pipelines/ls-grep/script-03.sh
TzeMingHo: So, if we only need the first character in upper case and no more upper case follows, can we use '^[A-Z][^A-Z]*$' ?

SlideGauge: Yes

illicitonion:

PR #393, opened 2026-03-19
London | 26-SDC-Mar | Craig D'Silva | Sprint 2 | jq

craig-dsilva:
## Learners, PR Template

Self checklist

## Changelist

I have completed the `jq` exercises.

### DIFF
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+
+cat person.json | jq -r ".name"
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -6,4 +6,4 @@ set -euo pipefail
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
 
-jq .name person.json
+cat person.json | jq -r ".name"
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -6,4 +6,4 @@ set -euo pipefail
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
 
-cat person.json | jq -r ".name"
+jq -r .name person.json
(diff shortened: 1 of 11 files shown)


jq/script-01.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+
+cat person.json | jq -r ".name"
```
cjyuan: Why not specify the input filename directly as:
```
jq -r .name person.json
```
?

jq/script-03.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person, then a comma, then their profession.
 # Your output should be exactly the string "Selma, Software Engineer", but should not contain any quote characters.
+
+cat person.json | jq -j '[.name, .profession] | join (", ")'
```
cjyuan: Why use `-j` option on this particular command?

cjyuan (commented):
Solution works. I just have a few questions.

jq/script-01.sh
craig-dsilva: I thought `cat` with pipelines was required. Didn't know I could use it directly, just like grep.

jq/script-03.sh
craig-dsilva: I wanted to get a raw output to match the terminal format.

cjyuan: `-j` option means "without adding a newline at the end". 

What do you mean by "get a raw output to match the terminal format"?

jq/script-01.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+
+jq .name person.json
```
cjyuan: Why remove the `-r` option? Now the requirement `should not contain any quote characters` is no longer satisfied.

jq/script-03.sh
craig-dsilva: Sorry, I used this to prevent a new line after output.

cjyuan:
Looks good.

PR #395, opened 2026-03-20
London | 26-SDC-Mar | Craig D'Silva | Sprint 2 | Shell Pipelines

craig-dsilva:
## Learners, PR Template

Self checklist

## Changelist

This is the coursework for the Shell pipelines exercise


shell-pipelines/sort-uniq-head-tail/script-06.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+
+cat events.txt | sort | uniq -c
```
SlideGauge: It groups by the whole line (person + event type), so the output is 2 Entry German, 1 Entry Mariana, etc. - you have to sum those up yourself to get 5 and 4. The task says it should be clear from the output that there are 5 Entry and 4 Exit events.

shell-pipelines/sort-uniq-head-tail/script-07.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
 # The word "Event" should not appear in your script's output.
+
+cat events-with-timestamps.txt | awk '{print $3, $4;}' | grep -v 'Event' | sort | uniq -c
```
SlideGauge: This approach will use awk to extract both event and name, thus leading to similar problem as script-06.sh

shell-pipelines/sort-uniq-head-tail/script-02.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
 # The first line of your output should be "Basia London 22 9 6" (with no quotes).
+
+sort -k3 -n -r scores-table.txt
```
SlideGauge: What does -k3 without specifying the end key mean and is it what we want here?

shell-pipelines/sort-uniq-head-tail/script-03.sh
```diff
@@ -8,3 +8,5 @@ set -euo pipefail
 # Basia London 22 9 6
 # Piotr Glasgow 15 2 25 11 8
 # Chandra Birmingham 12 6
+
+sort -k3 -n -r scores-table.txt | head -n 3
```
SlideGauge: What does -k3 without specifying the end key mean and is it what we want here?

shell-pipelines/sort-uniq-head-tail/script-04.sh
```diff
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with shows the line for the player whose first score was the second highest.
 # Your output should be: "Piotr Glasgow 15 2 25 11 8" (without quotes).
+
+sort -k3 -n -r scores-table.txt | head -n 2 | tail -n 1
```
SlideGauge: What does -k3 without specifying the end key mean and is it what we want here?

shell-pipelines/ls-grep/script-03.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should contain 7 files.
+
+ls sample-files | grep '^[A-Z]' | grep -vE '.[A-Z]'
```
SlideGauge: Creative use of piping of greps, like.

SlideGauge (commented):
Could you add trailing new lines to all the scripts? Doublecheck the settings of your editor, so it will place them automatically.

SlideGauge:
Could you apply fixes to the points I described?

shell-pipelines/sort-uniq-head-tail/script-02.sh
craig-dsilva: The `-k` flag means to sort by key.  By adding 3 to the flag, I'm telling the sort program to sort by the third column.

SlideGauge: Thanks for explaining what -k3 does — you're right that it selects a sort key. But I was asking something slightly narrower: with -k3 and no end specified, where does that key actually stop? Is it just field 3, or field 3 all the way to the end of the line? Try sort -k3 vs sort -k3,3 on a couple of rows and watch what changes. It happens to give the right answer on this data — but would it still hold if two people had the same first score?

shell-pipelines/sort-uniq-head-tail/script-07.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
 # The word "Event" should not appear in your script's output.
+
+cat events.txt | awk '{print $1}' | sort | uniq -c | awk '{print $1}'
```
SlideGauge, line 10: This is closer, but two things to check before I mark it resolved.
1. Which file are you reading? Re-read the TODO at the top of this script and compare it to the cat in your pipeline — are they the same file? There's a reason 07 points at a different input than 06; what's different about that file, and why might it matter here?
2. Yhe task says it should be clear that there are 5 Entry and 4 Exit events — can you still tell which number is which after your final step? What is that last awk '{print $1}' removing, and do you want it gone?
Also keep an eye on the "the word Event should not appear" requirement once you switch inputs — that constraint is a hint about what the correct file looks like.

SlideGauge (commented):
Better now, a couple of notes from my sides though

PR #397, opened 2026-03-20
London | 26-SDC-Mar | Ebrahim Beiati-Asl | Sprint 4 | Implement shell python shell tools

ebrahimbeiati:
## Self checklist

## Changelist

Writing ls.py, cat.py, wc.py

### DIFF
--- a/implement-shell-tools/implement-shell-tools-python/cat/cat.py
+++ b/implement-shell-tools/implement-shell-tools-python/cat/cat.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+import sys
+import os
+
+global_line_counter = 1
+
+def print_file(file_path, options):
+    global global_line_counter
+    try:
+        with open(file_path, "r", encoding="utf-8") as f:
+            content = f.read()
+
+        lines = content.split("\n")
+        if lines and lines[-1] == "":
+            lines.pop()
+
+        for line in lines:
+            prefix = ""
+
+            should_number = (
+                options["number_mode"] == "all" or
+                (options["number_mode"] == "non-empty" and line.strip() != "")
+            )
+
+            if should_number:
+                prefix = f"{global_line_counter:6}\t"
+                global_line_counter += 1
+
+            sys.stdout.write(prefix + line + "\n")
+
+    except Exception as e:
+        print(f"cat: {file_path}: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
+def main():
+    global global_line_counter
+    args = sys.argv[1:]
+
+    options = {"number_mode": "off"}
+    files = []
+
+    for arg in args:
+        if arg == "-n":
+            options["number_mode"] = "all"
+        elif arg == "-b":
+            options["number_mode"] = "non-empty"
+        else:
+            files.append(arg)
+
+    if not files:
+        print("cat: missing file operand", file=sys.stderr)
+        sys.exit(1)
+
+    for file_path in files:
+        global_line_counter = 1
+        print_file(file_path, options)
+
+
+if __name__ == "__main__":
+    main()
--- a/implement-shell-tools/implement-shell-tools-python/cat/cat.py
+++ b/implement-shell-tools/implement-shell-tools-python/cat/cat.py
@@ -1,60 +1,52 @@
 #!/usr/bin/env python3
 import sys
-import os
+import argparse
 
-global_line_counter = 1
 
-def print_file(file_path, options):
-    global global_line_counter
+def print_file(file_path, options, counter):
     try:
         with open(file_path, "r", encoding="utf-8") as f:
-            content = f.read()
+            for line in f:
+                stripped = line.rstrip("\n")
 
-        lines = content.split("\n")
-        if lines and lines[-1] == "":
-            lines.pop()
+                should_number = options.number_mode == "all" or (
+                    options.number_mode == "non-empty" and stripped != ""
+                )
 
-        for line in lines:
-            prefix = ""
+                if should_number:
+                    print(f"{counter}\t{stripped}")
+                    counter += 1
+                else:
+                    print(stripped)
 
-            should_number = (
-                options["number_mode"] == "all" or
-                (options["number_mode"] == "non-empty" and line.strip() != "")
-            )
+    except FileNotFoundError:
+        print(f"cat: {file_path}: No such file or directory", file=sys.stderr)
 
-            if should_number:
-                prefix = f"{global_line_counter:6}\t"
-                global_line_counter += 1
-
-            sys.stdout.write(prefix + line + "\n")
-
-    except Exception as e:
-        print(f"cat: {file_path}: {e}", file=sys.stderr)
-        sys.exit(1)
+    return counter
 
 
 def main():
-    global global_line_counter
-    args = sys.argv[1:]
+    parser = argparse.ArgumentParser(description="Concatenate files and print output")
+    parser.add_argument("-n", "--number", action="store_true", help="number all lines")
 
-    options = {"number_mode": "off"}
-    files = []
+    parser.add_argument(
+        "-b", "--number-nonblank", action="store_true", help="number non-empty lines"
+    )
 
-    for arg in args:
-        if arg == "-n":
-            options["number_mode"] = "all"
-        elif arg == "-b":
-            options["number_mode"] = "non-empty"
-        else:
-            files.append(arg)
+    parser.add_argument("files", nargs="+", help="files to read")
 
-    if not files:
-        print("cat: missing file operand", file=sys.stderr)
-        sys.exit(1)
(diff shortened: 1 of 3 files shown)


implement-shell-tools/implement-shell-tools-python/cat/cat.py
```diff
@@ -0,0 +1,61 @@
...
+        lines = content.split("\n")
+        if lines and lines[-1] == "":
+            lines.pop()
+
+        for line in lines:
+            prefix = ""
+
+            should_number = (
+                options["number_mode"] == "all" or
+                (options["number_mode"] == "non-empty" and line.strip() != "")
```
Poonam-raj: Nice check here for whether the line should be counted.

implement-shell-tools/implement-shell-tools-python/cat/cat.py
```diff
@@ -0,0 +1,61 @@
...
+            options["number_mode"] = "non-empty"
+        else:
+            files.append(arg)
+
+    if not files:
+        print("cat: missing file operand", file=sys.stderr)
+        sys.exit(1)
+
+    for file_path in files:
+        global_line_counter = 1
```
Poonam-raj: This will reset the counter for every file but this isn't a behaviour for the cat command - we should be continuing the count across files.

implement-shell-tools/implement-shell-tools-python/cat/cat.py
```diff
@@ -0,0 +1,61 @@
...
+        if arg == "-n":
+            options["number_mode"] = "all"
+        elif arg == "-b":
+            options["number_mode"] = "non-empty"
+        else:
+            files.append(arg)
+
+    if not files:
+        print("cat: missing file operand", file=sys.stderr)
+        sys.exit(1)
```
Poonam-raj: I would consider using `argparse` to parse your arguments for the cat command. It allows you to set what you expect to receive. It also makes your code for the command at a lot clearer.

It does the hard work you're doing by printing a command error on line 52 for you as all you would need to do is set the expected arguments for the command. It allows you to add help information and a description of the command so if people need that when they're using `cat` they can use `--help`

Doing it manually like this can bake in some unexpected bugs if someone uses the command wrong.

implement-shell-tools/implement-shell-tools-python/cat/cat.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+import sys
+import os
+
+global_line_counter = 1
+
+def print_file(file_path, options):
+    global global_line_counter
```
Poonam-raj: I probably wouldn't use a global variable as a counter here. Could you think of ways you could make your functions a bit purer - relying less on a global variable and more on a variable that can be passed through your functions maybe?

implement-shell-tools/implement-shell-tools-python/ls/ls.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+import sys
+import os
+
+def list_directory(path, show_all, one_per_line):
```
Poonam-raj: `one_per_line` is not used in the `list_directory` function, unnecessary arg

implement-shell-tools/implement-shell-tools-python/ls/ls.py
```diff
@@ -0,0 +1,47 @@
...
+    except Exception as e:
+        print(f"Error reading directory {path}: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
+def main():
+    args = sys.argv[1:]
+
+    show_all = "-a" in args
+    one_per_line = "-1" in args
```
Poonam-raj: I think largely this `one_per_line` variable isn't used anywhere in the logic. It could've been omitted and still met the requirements.

implement-shell-tools/implement-shell-tools-python/ls/ls.py
```diff
@@ -0,0 +1,47 @@
...
+        for e in entries:
+            print(e)
+
+    except Exception as e:
+        print(f"Error reading directory {path}: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
+def main():
+    args = sys.argv[1:]
```
Poonam-raj: Again here, using `argparse` python lib would let you outline the parameters of the command more clearly in the code (helpful for others reading) and add ability for `--help` flag, and allow you to grab paths/args a lot easier with argparse doing the error handling for you

implement-shell-tools/implement-shell-tools-python/ls/ls.py
```diff
@@ -0,0 +1,47 @@
...
+            entries = [e for e in entries if not e.startswith(".")]
+
+        entries.sort(key=lambda x: (x not in (".", ".."), x))
+
+        for e in entries:
+            print(e)
+
+    except Exception as e:
+        print(f"Error reading directory {path}: {e}", file=sys.stderr)
+        sys.exit(1)
```
Poonam-raj, line 13: nice error message here

implement-shell-tools/implement-shell-tools-python/ls/ls.py
```diff
@@ -0,0 +1,47 @@
...
+    if not dirs:
+        dirs = [os.getcwd()]
+
+    for i, path in enumerate(dirs):
+            if os.path.isdir(path):
+                if len(dirs) > 1:
+                    print(f"{path}:")
+                list_directory(path, show_all, one_per_line)
+                if len(dirs) > 1 and i < len(dirs) - 1:
+                    print("")
```
Poonam-raj: only adding blank line between directories is a nice touch

implement-shell-tools/implement-shell-tools-python/wc/wc.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
+import sys
+
+def count_file_content(content):
+    raw_lines = content.split("\n")
+    lines = len(raw_lines) - 1 if raw_lines[-1] == "" else len(raw_lines)
+    words = len(content.split())
+    bytes_ = len(content.encode("utf-8"))
```
Poonam-raj, line 8: Nice inclusion of encoding to count bytes accurately

implement-shell-tools/implement-shell-tools-python/wc/wc.py
```diff
@@ -0,0 +1,70 @@
...
+    return lines, words, bytes_
+
+
+def print_counts(path, counts, options):
```
Poonam-raj, line 12: I like the separation of concerns with these functions

implement-shell-tools/implement-shell-tools-python/wc/wc.py
```diff
@@ -0,0 +1,70 @@
...
+            options["byte"] = True
+        else:
+            files.append(arg)
+
+    if not files:
+        print("No files specified", file=sys.stderr)
+        sys.exit(1)
+
+    if not any(options.values()):
+        options = {"line": True, "word": True, "byte": True}
```
Poonam-raj, line 46: Again I would say using argparse would take some of the heavy lifting from validating the extracted arguments and sending errors for file etc.

Poonam-raj (commented):
Good work here especially the `wc` command. Overall I would love to see you amend some bits before this is complete:
- Can you use argparse library to better manage your arguments across all of these commands?
- Remove unused one_per_line from ls command if you don't have that behaviour as optional
- fix the bug that resets the line count after each file in the cat command
- change line_counter variable in cat command so it isn't a global command

implement-shell-tools/implement-shell-tools-python/ls/ls.py
```diff
@@ -0,0 +1,39 @@
...
+    for entry in sorted(entries):
+        if not show_all and entry.startswith("."):
+            continue
+        print(entry)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="List directory contents")
+    parser.add_argument("-a", "--all", action="store_true", help="show hidden files")
+    parser.add_argument("paths", nargs="*", default=["."], help="directories to list")
```
Poonam-raj, line 26: Even if the core functionality will always act like `-1` has been passed in, i think we should still take the argument here so `ls` can be used as described in the instructions

implement-shell-tools/implement-shell-tools-python/wc/wc.py
Poonam-raj: Still would like to see a refactor to argparse here if you can

Poonam-raj (commented):
Just two areas where things need tweaking:

- argparse use for `wc` command
- missing argument for `ls` so we can use it as instructed even if extra functionality isn't needed in the logic

Poonam-raj (commented):
Looks great

illicitonion:

PR #403, opened 2026-03-20
Birmingham | 26-SDC-Mar | Chioma Okeke | Sprint 2 | jq

JanefrancessC:
## Learners, PR Template

Self checklist

## Changelist

Completed exercises in `/jq`.

### DIFF
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+jq -r '.name' < person.json
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -5,3 +5,5 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+# jq -r '.name' < person.json //using redirection to the input
+jq -r '.name' person.json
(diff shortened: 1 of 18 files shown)


jq/script-01.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+jq -r '.name' < person.json
```
cjyuan: Why not specify the file input directly as 
```
jq -r '.name' person.json
```
That is, without using input redirection?

jq/script-02.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the address of the person, all on one line, with a comma between each line.
 # Your output should be exactly the string "35 Fashion Street, London, E1 6PX", but should not contain any quote characters.
+jq -jr '.address | join(", "), "\n"' < ./person.json
```
cjyuan: Why use `-j` option and then manually output the newline character? (Your commands in other script don't follow this approach)

cjyuan (commented):
Your solution works. I just have a few questions.

jq/script-01.sh
JanefrancessC: I used redirection as a reminder of different approaches I can use to solve the problem.

jq/script-02.sh
JanefrancessC: Read about the -j flag as well, just a reminder that it takes out `\n' characters.

jq/script-01.sh
cjyuan: You could leave a note in the script and even include in the note all possible alternatives and their pros and cons.

JanefrancessC: I have modified all the tasks to call the file directly.

JanefrancessC:
> Your solution works. I just have a few questions.

@cjyuan  I have refactored the scripts to use the file directly rather than redirection

cjyuan:
Looks good.

PR #418, opened 2026-03-23
Glasgow | 26- SDC-Mar | Taras Mykytiuk | Sprint 2 | Shell pipelines

TarasMykytiuk:
## Learners, PR Template

Self checklist

## Changelist

Shell pipelines exercises are done.

### DIFF
--- a/shell-pipelines/ls-grep/script-02.sh
+++ b/shell-pipelines/ls-grep/script-02.sh
@@ -3,4 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter.
+ls sample-files | grep '\b[A-Z]'
 # Your output should contain 10 files.
--- a/shell-pipelines/ls-grep/script-02.sh
+++ b/shell-pipelines/ls-grep/script-02.sh
@@ -3,5 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter.
-ls sample-files | grep '\b[A-Z]'
+ls sample-files | grep '^[A-Z]'
 # Your output should contain 10 files.
--- a/shell-pipelines/ls-grep/script-03.sh
+++ b/shell-pipelines/ls-grep/script-03.sh
@@ -3,4 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
+ls sample-files | grep -w '[A-Z][a-z]*'
 # Your output should contain 7 files.
--- a/shell-pipelines/ls-grep/script-03.sh
+++ b/shell-pipelines/ls-grep/script-03.sh
@@ -3,5 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
-ls sample-files | grep -w '[A-Z][a-z]*'
+ls sample-files | grep '^[A-Z][a-z]*$'
 # Your output should contain 7 files.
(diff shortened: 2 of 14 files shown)


shell-pipelines/ls-grep/script-02.sh
```diff
@@ -3,4 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter.
+ls sample-files | grep '\b[A-Z]'
```
cjyuan: `\b` won't work for filenames like `abc.Xyz` because `.` is considered a word boundary.

Hint:  There is a regex symbol that matches "beginning of a string".

shell-pipelines/ls-grep/script-03.sh
```diff
@@ -3,4 +3,5 @@
 set -euo pipefail
 
 # TODO: Write a command to output the names of the files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
+ls sample-files | grep -w '[A-Z][a-z]*'
```
cjyuan: - Filenames are listed one filename per line. `-w` option does not help.

- Also, the regex should also work for filenames like `A123.jpg`, `B$$$.txt`.

Note: The same also apply to `script-04.sh`

shell-pipelines/sort-uniq-head-tail/script-06.sh
```diff
@@ -4,4 +4,6 @@ set -euo pipefail
 
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
+cat events.txt | grep 'Entry' | sort -u | wc -l | sed 's/$/ entries/'
+cat events.txt | grep 'Exit' | sort -u | wc -l | sed 's/$/ exits/'
```
cjyuan: Can you use only ONE command to produce an output in the form:
```
5 Entry
4 Exit
```
?

Hint: Extract the words in the first column first.

The same applies also to `script-07.sh`.

shell-pipelines/sort-uniq-head-tail/script-02.sh
```diff
@@ -4,4 +4,5 @@ set -euo pipefail
 
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with lines sorted by the person's first score, descending.
+sort scores-table.txt -k3 -n -r
```
cjyuan: Common practice is to specify options before the filename.

TarasMykytiuk: Thanks for hint.

shell-pipelines/sort-uniq-head-tail/script-06.sh
TarasMykytiuk: Fixed.

shell-pipelines/ls-grep/script-03.sh
TarasMykytiuk: Fixed.

shell-pipelines/ls-grep/script-02.sh
TarasMykytiuk: Fixed.

shell-pipelines/ls-grep/script-03.sh
cjyuan: Can the regex detect filenames like `A123` or `B$$$`?

shell-pipelines/ls-grep/script-02.sh
cjyuan: What do you mean by "FixeD"? I don't notice any change made on this file.

TarasMykytiuk: Somehow my changes was reversed before commit

shell-pipelines/ls-grep/script-03.sh
TarasMykytiuk: Now should.

cjyuan:
Changes are good.

illicitonion:

PR #437, opened 2026-03-26
London | 26-SDC-MAR | Jamal Laqdiem | Sprint 3 | Implement Shell Tools

jamallaqdiem:
## Learners, PR Template

Self checklist

## Changelist

shell tools exercises.


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,40 @@
...
+        } else {
+          // print empty lines
+          console.log(line);
+        }
+      } else {
+        console.log(line);
+      }
+    });
+  } catch (error) {
+    console.log(`Could not read: ${path}`);
```
illicitonion: If there's an error where will this message get printed? stdout or stderr? And what exit code would the process exit with?

Do these match our expectations from shell tools?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,40 @@
...
+      } else if (showNonBlankNumbers) {
+        // increment and show numbers only if the line is not empty.
+        if (line.trim() !== "") {
+          console.log(`${counterLines++}  ${line}`);
+        } else {
+          // print empty lines
+          console.log(line);
+        }
+      } else {
+        console.log(line);
```
illicitonion: The three branches here look quite similar and repetitive. In general, if you have multiple similar branches, it's more clear to extract the differences into variables, and then run the same code, i.e. so you'd only have one call to `console.log` which looks more like ``console.log(`${prefix}${line}\n`)`` where `prefix` may be set differently based on options (including potentially an empty string). 

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,40 @@
...
+
+// THis will give an array without the path to node and to the file.
+const argv = process.argv.slice(2);
+
+//Get line numbers.
+const showNumbers = argv.includes("-n");
+const showNonBlankNumbers = argv.includes("-b");
+
+//filter the - from the array argv as it's a flag.
+const filePaths = argv.filter((arg) => !arg.startsWith("-"));
```
illicitonion, line 9: What would happen if someone passed a `-q` flag here? What _should_ happen?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,28 @@
...
+
+// This will give an array without the path to node and to the file.
+const argv = process.argv.slice(2);
+
+// filter the flag to find the target folder.
+const filePaths = argv.filter((arg) => !arg.startsWith("-"));
+const showHiddenFiles = argv.includes("-a");
+
+// if no folder provide we use the current one
+const target = filePaths[0] || ".";
```
illicitonion, line 14: What would happen if someone specified multiple paths, e.g. `ls /some/file /some/other/file`? What does your implementation do?

The README.md only requires that your programme works with simple paths, but I would recommend implementing support for multiple. But if you _don't_ implement that, you generally want to give an error to the user if they supply input you don't expect, rather than just ignoring it.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,40 @@
...
+const noFlags = !showLines && !showCharacters && !showWords && !showCharacters;
+if (!filePaths) {
+  console.error("PLease provide a file path");
+  process.exit(1);
+}
+// loop trough the array to get each file path.
+filePaths.forEach((filePath) => {
+  const content = fs.readFileSync(filePath, "utf-8");
+
+  const lines = content.split("\n").length - 1;
```
illicitonion: Why do you have to `- 1` here?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,40 @@
...
+    .trim()
+    .split(/\s+/)
+    .filter((word) => word != "").length;
+  // here I used Buffer.byteLength even if characters and bytes can be the same number .length, however sometimes an emoji or special characters can be heavier 2b or 4b
+  const bytes = Buffer.byteLength(content);
+  const characters = content.length;
+
+  let output = "";
+
+  if (showLines || noFlags) output += `${lines} `;
```
illicitonion: Can you think of a way to achieve this same result, but without needing a `noFlags` variable at all?

If our program was bigger and more complicated, in most of the program we don't want to have to think about what the exact flags interface was. Ideally we can just look at `showLines` to decide whether we should show lines, without needing to worry which exactly flags led to us needing to show lines.

implement-shell-tools/cat/cat.js
jamallaqdiem: Thanks for reviewing, I followed your correct instructions , as it's much clearer and readable using a variable.

implement-shell-tools/cat/cat.js
jamallaqdiem: Thanks for reviewing,  I made the changes from console.log that wil send txt to stdout To console.error that willsend to stderr and display the error in the screen.
Added pocess.exit(1) to make sure that in case of failure the code should exit with a non  zero code.

implement-shell-tools/cat/cat.js
jamallaqdiem: Thanks for reviewing, you are right to point that out, my logic will let the user confused thinking that the -q flag exists and worked, I implemented in place a flag check and ensured print an error and exit with not 0 code.

implement-shell-tools/ls/ls.js
jamallaqdiem: Thanks for reviewing, I should at least display an stderr and exit .

implement-shell-tools/wc/wc.js
jamallaqdiem: Thanks for reviewing, I used  -1 because my vscode automatically add a new line when save, so in this case split method will add an empty element, however in other environments this logic my fail if no trialling new  line added .

jamallaqdiem: We may add a conditional check using .pop method in case of empty string at the end

implement-shell-tools/wc/wc.js
jamallaqdiem: Thanks for reviewing, I implemented a clear way to switch all flags to true if no flag was picked.
Fixed  a bug.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,48 @@
+import process from "node:process";
+import { promises as fs } from "node:fs";
+
+// THis will give an array without the path to node and to the file.
+const argv = process.argv.slice(2);
+
+//Get line numbers.
```
illicitonion: Does this comment provide value beyond the variable names?

If yes, could you change the variable names to make this comment no longer necessary?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,48 @@
...
+const showNonBlankNumbers = argv.includes("-b");
+
+//filter the - from the array argv as it's a flag.
+const filePaths = argv.filter((arg) => !arg.startsWith("-"));
+
+const flagsUsed = argv.filter((arg) => arg.startsWith("-"));
+const supportedFlags = ["-n", "-b"];
+for (const flag of flagsUsed) {
+  if (!supportedFlags.includes(flag)) {
+    console.error(`Invalid option try 'node cat.js --help' for more info.`);
```
illicitonion: What does `node cat.js --help` do? As far as I can tell, it will probably print this error?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,48 @@
...
+        if (line.trim() !== "") {
+          prefix = `${counterLines++}  `;
+        }
+      } else if (showNumbers) {
+        prefix = `${counterLines++}  `;
+      }
+      console.log(`${prefix}${line}`);
+    });
+  } catch (error) {
+    console.error(`Could not read: ${path}`);
```
illicitonion: Right now you're always printing the same error message if something went wrong.

What are the advantages/disadvantages to including the specific error message in the output to the user?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,32 @@
...
+if (filePaths.length > 1) {
+  console.error("Error: This version only supports one directory path a time.");
+  process.exit(1);
+}
+// if no folder provide we use the current one
+const target = filePaths[0] || ".";
+// read the file.
+const files = fs.readdirSync(target);
+
+// save the result into the variable.
```
illicitonion: Almost all of your comments provide no value over just reading the code. Comments should explain _why_ we're doing something (if it's not obvious), or explain something particularly confusing/tricky. If you can read the code, the comments here aren't useful - please consider all of them and remove any you think aren't useful.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,46 @@
...
+// if no flags enable all.
+if (!showLines && !showCharacters && !showWords && !showBytes) {
+  showLines = true;
+  showCharacters = true;
+  showWords = true;
+  showBytes = true;
+}
+// fix bug .length ===0 will be true if nothing provided, instead !filePath will return empty array which result true.
+if (filePaths.length === 0) {
+  console.error("PLease provide a file path");
```
illicitonion, line 21: You frequently have typos where the second letter after a capital letter is also capitalised. Please look out for these and fix them.

implement-shell-tools/wc/wc.js
illicitonion: Please do the conditional pop check :)

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,46 @@
...
+
+  let output = "";
+
+  if (showLines) output += `${lines} `;
+  if (showWords) output += `${words} `;
+  if (showBytes) output += `${bytes} `;
+  if (showCharacters) output += `${characters} `;
+
+  console.log(`${output} ${filePath}`);
+});
```
illicitonion: Your output doesn't match real `wc`'s output if multiple files were passed. Please make sure to run the examples given in the READMEs and that your behaviours match.

implement-shell-tools/cat/cat.js
jamallaqdiem: The advantages is that will show the users the error and allow him to adjusted themselves.
The disadvantages is that it may leak sensitive data like a name in the path.

jamallaqdiem:
Thanks for the review, I addressed all matters.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,44 @@
...
+    splitLines.forEach((line) => {
+      let prefix = "";
+
+      if (showNonBlankNumbers) {
+        if (line.trim() !== "") {
+          prefix = `${counterLines++}  `;
+        }
+      } else if (showNumbers) {
+        prefix = `${counterLines++}  `;
+      }
```
illicitonion, line 37: I would think about collapsing this into one if:
```js
if (showNumbers || (howNonBlankNumbers && line.trim() !== "")) {
```

because they have the same intent, but wouldn't push strongly for this if you prefer it as-is :)

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,27 @@
...
+if (filePaths.length > 1) {
+  console.error("Error: This version only supports one directory path a time.");
+  process.exit(1);
+}
+
+const target = filePaths[0] || ".";
+
+const files = fs.readdirSync(target);
+
+let filteredFIles = files;
```
illicitonion, line 18:
```suggestion
let filteredFiles = files;
```

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,67 @@
...
+
+    console.log(`${output} ${filePath}`); 
+  } catch(error) {
+    console.error(`Error reading ${filePath}: ${error.message}`);
+  }
+}); 
+
+if (filePaths.length > 1) {
+  let totalOutput = "";
+  if (showLines)  totalOutput += `${totalLines.toString().padStart(8)}`;
```
illicitonion, line 60: This is a bit repetitive with the single-file formatting case :)

implement-shell-tools/cat/cat.js
illicitonion: Yeah, this is generally correct, though specific on names in the path, the user presumably has access to the name in the path because if they didn't they wouldn't have been able to pass it as an argument.

illicitonion (commented):
Looks much better, well done!

illicitonion:

PR #440, opened 2026-03-27
London | 26-SDC-Mar | Mariia Serhiienko | Sprint 2 | Shell pipelines

Konvaly:
## Learners, PR Template

Self checklist

## Changelist

Completed tasks with commands ls-grep-sort-uniq-grep-tail-tr

Task ID: CYF-1149

### DIFF
--- a/shell-pipelines/sort-uniq-head-tail/script-04.sh
+++ b/shell-pipelines/sort-uniq-head-tail/script-04.sh
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with shows the line for the player whose first score was the second highest.
 # Your output should be: "Piotr Glasgow 15 2 25 11 8" (without quotes).
+sort -k3 -rn scores-table.txt | head -2 | tail -1
--- a/shell-pipelines/sort-uniq-head-tail/script-06.sh
+++ b/shell-pipelines/sort-uniq-head-tail/script-06.sh
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+sort events.txt | uniq -c | sort -k2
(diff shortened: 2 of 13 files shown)


shell-pipelines/sort-uniq-head-tail/script-04.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the scores-table.txt file.
 # TODO: Write a command to output scores-table.txt, with shows the line for the player whose first score was the second highest.
 # Your output should be: "Piotr Glasgow 15 2 25 11 8" (without quotes).
+sort -k3 -rn scores-table.txt | head -2 | tail -1
```
cjyuan, line 8: Note: 
I was curious and asked AI's which of the following ways of using `head` and `tail` is preferred:
1) `head -2 | tail -1`
2) `head -n 2 | tail -n 1`

They suggested the 2nd syntax for compatibility and readability reasons.

shell-pipelines/sort-uniq-head-tail/script-06.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+sort events.txt | uniq -c | sort -k2
```
cjyuan, line 8: The output produced by your command does not quite match the expected output, which is in the form:
```
5 Entry
4 Exit
```

shell-pipelines/sort-uniq-head-tail/script-07.sh
```diff
@@ -6,3 +6,4 @@ set -euo pipefail
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
 # The word "Event" should not appear in your script's output.
+tail -n +2 events-with-timestamps.txt | cut -d' ' -f3 | sort | uniq -c
```
cjyuan, line 9: `cut -d' ' -f3 works only if the fields on each line is separated by EXACTLY one space.
The fields in the given file are separated by a tab key, so the command won't work.

PR #465, opened 2026-04-04
London | 26-SDC-March | Zobeir Rigi| Sprint 2 | jq

Zobeir-Rigi:
Self checklist

## Changelist

Jq practice scripts for learning data extraction and filtering.

### DIFF
--- a/jq/script-01.sh
+++ b/jq/script-01.sh
@@ -5,4 +5,4 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
-jq -r  ."name" person.json
+    jq -r  ."name" person.json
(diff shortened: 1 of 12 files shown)


jq/script-01.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the person.json file.
 # TODO: Write a command to output the name of the person.
 # Your output should be exactly the string "Selma", but should not contain any quote characters.
+    jq -r  ."name" person.json
```
cjyuan, line 8: -Indentation is off

- Why place double quotes around `name`?

cjyuan (commented):
Solution looks good. I just have one question.

jq/script-01.sh
Zobeir-Rigi: I used quotes because jq allows object keys to be written as strings, but .name is the cleaner shorthand here.

illicitonion:

PR #472, opened 2026-04-07
Automate testing of initial sprints in tools module

LonMcGregor:
* Expected output now in expected folder
* Github action which runs after validation, when Needs Review
* Script that can be run to test locally before push
* Split number sys qs into automatable/manualy checkable


.github/workflows/test-sdc-tasks.yml
```diff
@@ -0,0 +1,65 @@
...
+
+on:
+  pull_request:
+    types:
+      - opened
+      - labeled
+      - reopened
+      - synchronize
+    paths:
+      - '**.sh'
```
illicitonion: Does this not mean that it won't trigger for the number systems README.md file change?

.github/workflows/test-sdc-tasks.yml
```diff
@@ -0,0 +1,65 @@
...
+jobs:
+  test_sdc_tasks:
+
+    name: Test SDC Tasks
+    runs-on: ubuntu-slim
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: checkout base branch
```
illicitonion: Shall we add a filtering step that cancels this run if there isn't a "Needs Review" label?

And maybe also add a dependency on the 'Validate PR Metadata' workflow so that we don't run if it's failing?

test-sdc.sh
```diff
@@ -0,0 +1,134 @@
...
+		fi
+		for exercise in *.sh; do
+			total=$(($total+1))
+			./$exercise > ../../test.tmp
+			cmp ../../test.tmp ../../expect/individual-shell-tools/$directory$exercise --quiet
+			if [ $? -eq 0 ]; then
+				pass=$(($pass+1))
+			else
+				if [[ "$exercise" == *"stretch"* ]]; then
+					# echo "Stretch task $directory$exercise failing, you can ignore this if you did not attempt. If you didnt, please have the volunteer check this."
```
illicitonion: Shall we check if the file was modified, and ignore if it wasn't?

test-sdc.sh
```diff
@@ -0,0 +1,134 @@
...
+					total=$(($total-1))
+				else
+					echo "Failed $directory$exercise, please either attempt again or have the volunteer check this." >> ../../testoutput.txt
```
illicitonion: "have the volunteer check this" - is this an instruction to the trainee ("Please find a volunteer") or to a volunteer ("This PR is in the Needs Review list for a reason") or something else?

test-sdc.sh
```diff
@@ -0,0 +1,134 @@
...
+		echo "attempted=y" >> "$GITHUB_OUTPUT"
+	fi
+	if [ $pass -eq $total ]; then
+		echo "This task is complete!" >> testoutput.txt
+		if [ -v GITHUB_OUTPUT ]; then
+			echo "complete=y" >> "$GITHUB_OUTPUT"
+		fi
+	fi
+	cat testoutput.txt
+elif [[ "$1" == "number-systems" ]]; then
```
illicitonion, line 106: I'd be tempted to separate this out into two files? One automatically checked, one manually checked?

illicitonion (commented):
Generally looks really good, and very useful!

One thing I really like about the metadata validator one that this doesn't do is collapse old comments... I just put together https://github.com/CodeYourFuture/trainee-tracker/pull/68 so that we can just run a standalone binary to close existing comments - if we add `<!--CYFTT tag: sdc-test-results-->` somewhere in the comment we leave, we can use that binary to minimise past comments. WDYT?

LonMcGregor:
I think I've got it working now. I've tested a fork in [my repo](https://github.com/LonMcGregor/Module-Tools/pull/3) and it seems to be working well. But because the action setup is tricky I won't know if it actually fully works until it's merged and I can do a full run through of solving a task using a fork from this repo. One thing I'm not sure of is how this is going to impact people who have already forked this repo - when they submit, it should skip past any irrelevant tasks, but for the handful that still haven't submitted these early tasks it might cause a problem because their repo won't have the right files in it.

illicitonion (approved):
Let's ship it and find out!

I believe because we're using `pull_request` not `pull_request_target`, this shouldn't impact any PRs that don't include this commit, so we hopefully won't run into branches-out-of-sync issues 🤞 worst case we can revert until the cohort finishes.

Let's squash+merge this to do away with all the commits on the way :)

PR #481, opened 2026-04-09
London | 26-SDC-March | Zobeir Rigi| Sprint 2 | shell-pipelines

Zobeir-Rigi:
Self checklist

### DIFF
--- a/shell-pipelines/ls-grep/script-04.sh
+++ b/shell-pipelines/ls-grep/script-04.sh
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
+ls sample-files | grep -E '^[A-Z][^A-Z]*$' | wc -l
--- a/shell-pipelines/ls-grep/script-04.sh
+++ b/shell-pipelines/ls-grep/script-04.sh
@@ -4,4 +4,11 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
-ls sample-files | grep -E '^[A-Z][^A-Z]*$' | wc -l
+ ls sample-files | grep -c -E '^[A-Z][^A-Z]*$'
+# -c → tells grep to count matching lines instead of printing them
+
+
+# -E → enables extended regex (more flexible syntax)
+# ^[A-Z] → starts with uppercase
+# [^A-Z]* → no other uppercase letters allowed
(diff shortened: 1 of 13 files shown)


shell-pipelines/ls-grep/script-04.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to count the number of files in the sample-files directory whose name starts with an upper case letter and doesn't contain any other upper case letters.
 # Your output should be the number 7.
+ls sample-files | grep -E '^[A-Z][^A-Z]*$' | wc -l
```
LonMcGregor: Can this be solved using only ls and grep?

shell-pipelines/sort-uniq-head-tail/script-03.sh
LonMcGregor, line 1: Did you commit the right version of this file?

LonMcGregor (commented):
Good attempt, there are a couple of places you might need to look again

shell-pipelines/ls-grep/script-04.sh
Zobeir-Rigi: No, because ls and grep can filter and display matching filenames, but they cannot count them. To get the number of files, you need an additional tool like wc -l or another counting method.

shell-pipelines/sort-uniq-head-tail/script-03.sh
Zobeir-Rigi: I've checked, and it should be **sort scores-table.txt**

shell-pipelines/ls-grep/script-04.sh
LonMcGregor: Are there no options you could use with grep that do counting?

shell-pipelines/sort-uniq-head-tail/script-03.sh
LonMcGregor: Can you push an update to the file please?

shell-pipelines/ls-grep/script-04.sh
Zobeir-Rigi: Yes, we can also use
 **ls sample-files | grep -c -E '^[A-Z][^A-Z]*$'**
-c → tells grep to count matching lines instead of printing them

shell-pipelines/sort-uniq-head-tail/script-03.sh
Zobeir-Rigi: Changes committed and pushed.

shell-pipelines/ls-grep/script-04.sh
LonMcGregor: good answer! using just one commands does make it simpler

LonMcGregor:
Good work, this is complete now

illicitonion:

PR #485, opened 2026-04-10
London | 26-SDC-Mar | Khilola Rustamova| Sprint 3 |implementing shell tools 

HilolaRustam:
## Learners, PR Template

Self checklist

## Changelist

Implementing shell tools for cat, ls and wc is complete.

### DIFF
--- a/implement-shell-tools/cat/my-cat.js
+++ b/implement-shell-tools/cat/my-cat.js
@@ -1,0 +1,41 @@
+#!/usr/bin/env node
+
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
+let numberLines = false;
+let numberNonEmpty = false;
+
+if (args[0] === "-n") {
+  numberLines = true;
+  args.shift();
+} else if (args[0] === "-b") {
+  numberNonEmpty = true;
+  args.shift();
+}
+let count = 1;
+
+args.forEach((file) => {
+  const content = fs.readFileSync(file, "utf-8");
+  const lines = content.split(/\r?\n/);
+  
+
+  if (numberLines){ 
+    lines.forEach((line) => {
+      console.log(`${count} ${line}`);
+      count++;
+    });
+  } else if (numberNonEmpty){
+    lines.forEach((line) => {
+    if (line !== "") {
+      console.log(`${count} ${line}`);
+      count++;
+    } else {
+      console.log("");
+    }
+  });
+} else {
+console.log(content);
+  }
+});
--- a/implement-shell-tools/cat/my-cat.js
+++ b/implement-shell-tools/cat/my-cat.js
@@ -6,15 +6,17 @@ const args = process.argv.slice(2);
 
 let numberLines = false;
 let numberNonEmpty = false;
+let files = [];
 
-if (args[0] === "-n") {
+for (const arg of args) {
+if (arg === "-n") {
   numberLines = true;
-  args.shift();
-} else if (args[0] === "-b") {
+} else if (arg === "-b") {
   numberNonEmpty = true;
-  args.shift();
-}
-let count = 1;
+} else {
+  files.push(arg);
+}   
+};
 
 args.forEach((file) => {
   const content = fs.readFileSync(file, "utf-8");
--- a/implement-shell-tools/cat/my-cat.js
+++ b/implement-shell-tools/cat/my-cat.js
@@ -18,22 +18,26 @@ for (const arg of args) {
     files.push(arg);
   }
 }
+let count = 1;
 
 //Process each files
-args.forEach((file) => {
+files.forEach((file) => {
   const content = fs.readFileSync(file, "utf-8");
   const lines = content.split(/\r?\n/);
-
-  let count = 1;
+  if (lines[lines.length - 1] == ""){lines.pop();
+  }
 
   lines.forEach((line) => {
-    const shouldNumber = numberLines || (numberNonEmpty && line !== "");
-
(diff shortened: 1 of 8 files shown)


illicitonion (commented):
This is impressive (bash is hard!), but the exercise here was to implement the tools _in javascript_ not in bash - can you take another look? :)

implement-shell-tools/cat/my-cat.js
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
+let numberLines = false;
+let numberNonEmpty = false;
+
+if (args[0] === "-n") {
```
illicitonion: What would happen if someone passed both flags (e.g. `node my-cat.js -b -n /some/file`)? What do you think _should_ happen?

implement-shell-tools/cat/my-cat.js
```diff
@@ -0,0 +1,41 @@
...
+  args.shift();
+}
+let count = 1;
+
+args.forEach((file) => {
+  const content = fs.readFileSync(file, "utf-8");
+  const lines = content.split(/\r?\n/);
+  
+
+  if (numberLines){
```
illicitonion: This code is a bit repetitive - in both the `numberLines` case and the `numberNonEmpty` case you have the same code adding a prefix and incrementing the count.

Can you think how to avoid this, so we can see more easily what's the _same_ between each branch, and what the differences are?

implement-shell-tools/cat/my-cat.js
```diff
@@ -0,0 +1,41 @@
...
+    if (line !== "") {
+      console.log(`${count} ${line}`);
+      count++;
+    } else {
+      console.log("");
+    }
+  });
+} else {
+console.log(content);
+  }
```
illicitonion: These lines aren't indented correctly

implement-shell-tools/ls/my-ls.js
illicitonion, line 1: When I ran `ls -1 -a sample-files` vs `node my-ls.js -1 -a sample-files` I saw slightly different output - can you check what's causing the difference?

implement-shell-tools/wc/my-wc.js
```diff
@@ -0,0 +1,67 @@
...
+
+    if (flag === "-l") {
+      console.log(`${lines} ${file}`);
+    } else if (flag === "-w") {
+      console.log(`${words} ${file}`);
+    } else if (flag === "-c") {
+      console.log(`${chars} ${file}`);
+    } else {
+      console.log(`${lines} ${words} ${chars} ${file}`);
+    }
```
illicitonion: This works if you specify exactly one flag, but the real `wc` supports multiple flags at the same time, e.g. `wc -w -l /some/file` - can you support this too?

implement-shell-tools/wc/my-wc.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  if (files.length > 1) {
+    if (flag === "-l") {
```
illicitonion: This is quite repetitive with the per-file printing - can you avoid that duplication?

illicitonion (commented):
This is generally looking good, but a few things to look at :) Well done!

implement-shell-tools/ls/my-ls.js
```diff
@@ -0,0 +1,59 @@
...
+function parseArgs(args) {
+  let showAll = false;
+  let targetDir = ".";
+
+  for (const arg of args) {
+    if (arg === "-a") {
+      showAll = true;
+    } else if (!arg.startsWith("-")) {
+      targetDir = arg;
+    }
```
illicitonion: If I run this program like:

```
node my-ls.js -a -b -c /some/path
```

what will you do with the arguments `-b` and `-c`? Is this a good experience for the user?

illicitonion (commented):
This is generally looking good - one thing to think about

implement-shell-tools/ls/my-ls.js
HilolaRustam: My current implementation uses fs.readdirSync(), which returns directory entries in a flat list without guaranteeing the same ordering semantics as Unix ls.

fs.readdirSync() provides a higher-level abstraction than the Unix ls syscall, and does not include . and ... Therefore, matching ls -a exactly requires explicitly simulating these entries and enforcing ordering rules, rather than relying purely on the filesystem API output.

ls -a does not treat hidden files separately — it includes them in the same sorted output as normal files, with . and .. always appearing first.

Because of this, to match ls exactly, I would need to explicitly enforce the same ordering rules: sorting the full dataset uniformly and then applying special placement for . and ... Without that, the ordering (e.g. hidden files appearing first or in different positions) can differ from real ls.

implement-shell-tools/ls/my-ls.js
HilolaRustam: In my old implementation, -b and -c would be ignored because they are not explicitly handled and are treated as unknown flags. 

This is not ideal from a user experience perspective, because the program silently ignores input that the user may expect to have an effect. 

A better approach would be to either warn the user about unknown flags or throw an error, so that invalid or unsupported options are explicitly surfaced rather than ignored.

illicitonion (commented):
Marking this as complete because you've addressed all of the comments - there are a few things that could be improved, but we've been working in the Python version of this PR on those things. You may want to try applying those ideas to your JS version too, but I'm not going to force you to :)

PR #487, opened 2026-04-10
London | 26-Mar-SDC | Onur Atas | Sprint 2| Shell pipelines

onurat:
self checklist

## Changelist
Solved shell pipeline exercises using grep, sort, uniq, and tr


shell-pipelines/sort-uniq-head-tail/script-06.sh
```diff
@@ -5,3 +5,4 @@ set -euo pipefail
 # The input for this script is the events.txt file.
 # TODO: Write a command to show how many times anyone has entered and exited.
 # It should be clear from your script's output that there have been 5 Entry events and 4 Exit events.
+sort events.txt | uniq -c
```
LonMcGregor, line 8: Can you find any way to make it more obvious that there are 5 Entry and 4 Exit?

It should be possible to have this be the output, so the user doesn't have to add the numbers up as they are reading it.

LonMcGregor (commented):
Good work on this task, just one file to take a look at

PR #496, opened 2026-04-15
London | 26-SDC-MAR | Craig D'Silva | Sprint 5 | Prep

craig-dsilva:
## Learners, PR Template

Self checklist

## Changelist

This is the prep work for week 5.

### DIFF
--- a/week-5-prep/dataclasses_exercise.py
+++ b/week-5-prep/dataclasses_exercise.py
@@ -1,0 +1,14 @@
+from dataclasses import dataclass
+import datetime as dt
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: dt.date
+    preferred_operating_system: str
+
+imran = Person("Imran", "2004-04-18", "Ubuntu")
+print(imran)
+
+imran2 = Person("Imran", "2004-04-18", "Ubuntu")
+print(imran == imran2)
--- a/week-5-prep/methods_exercise.py
+++ b/week-5-prep/methods_exercise.py
@@ -1,0 +1,14 @@
+import datetime as dt
+
+class Person:
+    def __init__(self, name: str, dob: str, preferred_operating_system: str):
+        self.name = name
+        self.age = dt.date.today().year - int(dob.split("-")[0])
+        
+        self.preferred_operating_system = preferred_operating_system
+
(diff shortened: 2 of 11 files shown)


week-5-prep/methods_exercise.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Person:
+    def __init__(self, name: str, dob: str, preferred_operating_system: str):
+        self.name = name
+        self.age = dt.date.today().year - int(dob.split("-")[0])
+        
+        self.preferred_operating_system = preferred_operating_system
+
+    def is_adult(self):
+        return self.age >= 18
```
LonMcGregor, line 11: Can you think of any conditions where this approach of pre-calculating the age in the constructor might cause problems?

week-5-prep/dataclasses_exercise.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+import datetime as dt
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: dt.date
+    preferred_operating_system: str
```
LonMcGregor, line 8: Can you add the is_adult method here and test that it still works?

LonMcGregor (commented):
Good work, can you review my 2 comments please

week-5-prep/methods_exercise.py
craig-dsilva: No, I'm not sure.

week-5-prep/dataclasses_exercise.py
craig-dsilva: In this exercise, we need to add the date of birth instead of age. In order to check if the person is an adult, I would need to create another method which calculates the date of birth with today's date and return a boolean if the person is an adult.

LonMcGregor (commented):
While calculating the age at class instantiation time is not a wrong approach, if this were part of a long-lived application, it's possible that enough time could pass that what is stored in age would be out of date with how things actually are. I'm also not sure if this accurately captures the full age, as it only looks at whole years and not partial years.

Could you combine this with your suggestion for the other task, and try to build a method that can figure out whether they are an adult, using the dob, and whatever the current date is when the method is called?

PR #497, opened 2026-04-15
Birmingham | 2026-MAR-SDC | Joy Opachavalit | Sprint 4 | Implement shell tools in python

enjoy15:
## Learners, PR Template

Self checklist

## Changelist

Completed all exercises

### DIFF
--- a/implement-shell-tools/ls/ls.py
+++ b/implement-shell-tools/ls/ls.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+
+import locale
+import os
+import sys
+
+
+def parse_args(args):
+    one_per_line = False
+    show_all = False
+    paths = []
+
+    for arg in args:
+        if arg.startswith("-") and arg != "-":
+            for flag in arg[1:]:
+                if flag == "1":
+                    one_per_line = True
+                elif flag == "a":
+                    show_all = True
+                else:
+                    print(f"ls: invalid option -- '{flag}'", file=sys.stderr)
+                    sys.exit(1)
+        else:
+            paths.append(arg)
+
+    if not one_per_line:
+        print("Usage: ls.py -1 [-a] [path]", file=sys.stderr)
+        sys.exit(1)
+
+    if len(paths) > 1:
+        print("Usage: ls.py -1 [-a] [path]", file=sys.stderr)
+        sys.exit(1)
+
+    return show_all, (paths[0] if paths else ".")
+
+
+def list_entries(path, show_all):
+    try:
+        entries = os.listdir(path)
+    except FileNotFoundError:
+        print(f"ls: cannot access '{path}': No such file or directory", file=sys.stderr)
+        sys.exit(1)
+    except NotADirectoryError:
+        print(os.path.basename(path))
+        return
+
+    if show_all:
+        entries = [".", ".."] + entries
+    else:
+        entries = [name for name in entries if not name.startswith(".")]
+
+    locale.setlocale(locale.LC_COLLATE, "")
+    entries = sorted(entries, key=locale.strxfrm)
+
+    for entry in entries:
+        print(entry)
+
+
+def main():
+    show_all, path = parse_args(sys.argv[1:])
+    list_entries(path, show_all)
+
+
+if __name__ == "__main__":
+    main()
--- a/implement-shell-tools/ls/ls.py
+++ b/implement-shell-tools/ls/ls.py
@@ -29,8 +29,13 @@ def main():
         entries = [name for name in entries if not name.startswith(".")]
 
     locale.setlocale(locale.LC_COLLATE, "")
-    for entry in sorted(entries, key=locale.strxfrm):
-        print(entry)
+    sorted_entries = sorted(entries, key=locale.strxfrm)
+
+    if args.one_per_line:
+        for entry in sorted_entries:
+            print(entry)
+    else:
+        print("  ".join(sorted_entries))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,65 @@
...
+    except NotADirectoryError:
+        print(os.path.basename(path))
+        return
+
+    if show_all:
+        entries = [".", ".."] + entries
+    else:
+        entries = [name for name in entries if not name.startswith(".")]
+
+    locale.setlocale(locale.LC_COLLATE, "")
```
LonMcGregor, line 31: Good idea to consider how this might affect sorting.

LonMcGregor (commented):

Is there a package you could use to help with managing arguments? Right now it is a bit complex, and you can't combine single-letter arguments like the command line tools.

implement-shell-tools/ls/ls.py
enjoy15: Thanks! I switched argument parsing to argparse and added combined short-flag expansion, so -1a/-a1 work like standard tools.
I also adjusted the -a path so . and .. are included before sorting, keeping sort behavior consistent.

LonMcGregor: Did you remember to push the changes? I can't see that on the PR yet.

enjoy15: Sorry, I have now pushed the changes.

implement-shell-tools/ls/ls.py
```diff
@@ -1,37 +1,53 @@
...
 import locale
 import os
 import sys
 
 
-def parse_args(args):
-    one_per_line = False
-    show_all = False
-    paths = []
+class LsArgumentParser(argparse.ArgumentParser):
```
LonMcGregor: Is there a reason you create a custom subclass over argparse?

implement-shell-tools/ls/ls.py
```diff
@@ -1,37 +1,53 @@
...
-                if flag == "1":
-                    one_per_line = True
-                elif flag == "a":
-                    show_all = True
-                else:
+        if arg == "-":
+            expanded.append(arg)
+            continue
+
+        if arg.startswith("-") and not arg.startswith("--") and len(arg) > 2:
```
LonMcGregor: the argparse library can handle combined flags - is there a particular functionality you are trying to achieve here?

LonMcGregor (commented):
Good work switching to using the argparse library. I do have some questions, as it looks like you are maybe not using the library as effectively as you could be.

implement-shell-tools/ls/ls.py
enjoy15: Good point. I used a custom subclass mainly so I could control the error output and keep the messages closer to the expected ls style, but there isn’t any special parsing behavior that truly requires it.

implement-shell-tools/ls/ls.py
enjoy15: I was also trying to support combined short flags like -1a, but if argparse already handles that cleanly here, I’m happy to simplify the code and rely on the built-in behavior instead. That would definitely make the implementation clearer.

LonMcGregor:
Thanks for replying @enjoy15 - Could you try implementing the ls using the standard argparse without subclassing or adding custom functions? It would help to reduce the amount of code that you have written, making it more readable and maintainable

enjoy15:
@LonMcGregor  Suggestion implemented. Thank you!

implement-shell-tools/ls/ls.py
```diff
@@ -6,76 +6,35 @@
...
-
-    if not parsed.one_per_line:
-        print("Usage: ls.py -1 [-a] [path]", file=sys.stderr)
-        sys.exit(1)
-
-    return parsed.show_all, parsed.path
+    args = parser.parse_args()
 
+    if not args.one_per_line:
+        parser.error("the following arguments are required: -1")
```
LonMcGregor: Why is the -1 argument required?

LonMcGregor (commented):
Great, this looks a lot cleaner now. One question, see my comment

implement-shell-tools/ls/ls.py
enjoy15: Good point! I enforced -1 to match the README examples, but the real ls doesn't require it. I've removed that check. Now -1 is truly optional and the tool works either way. The default behavior is still one per line, matching standard ls.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,37 @@
...
+        return
+
+    if args.show_all:
+        entries = entries + [".", ".."]
+    else:
+        entries = [name for name in entries if not name.startswith(".")]
+
+    locale.setlocale(locale.LC_COLLATE, "")
+    for entry in sorted(entries, key=locale.strxfrm):
+        print(entry)
```
LonMcGregor: Think about the purpose of the -1 argument. How should the output differ when it is activated and when it is not? Does your implementation meet that expectation?

LonMcGregor (commented):
The -1 is no longer required, but I think you still need to think about how the -1 argument works, see my comment

implement-shell-tools/ls/ls.py
enjoy15: Thank you for the catch! I realized that my original implementation was always printing one item per line, meaning the -1 flag wasn't actually changing the behavior.

I have updated the code so that it now checks for the one_per_line argument. If it's active, it prints one entry per line; otherwise, it outputs the entries horizontally, separated by spaces, to mimic the default ls behavior.

LonMcGregor (commented):
Thanks, this ls implementation matches the spec now. well done

enjoy15:
> Thanks, this ls implementation matches the spec now. well done

Thank you!

illicitonion:

PR #499, opened 2026-04-17
London | 26-SDC-Mar | Mariia Serhiienko | Sprint 3 | Implement shell tools

Konvaly:
## Learners, PR Template

Self checklist

## Changelist

Implemented three shell tools in NodeJS:

- cat  - prints file contents, supports -n (number all lines) and -b (number non-blank lines only)
- ls - lists directory contents, supports -1 (one per line) and -a (show hidden files)
- wc - counts lines, words and bytes in files, supports -l, -w, -c flags and multiple files with a total row

Each tool matches the output of the real Unix commands.

Task ID: CYF-1150


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,54 @@
...
+import process from "node:process";
+
+const args = process.argv.slice(2);
+
+if (args.length === 0) {
+  console.error("Usage: node cat.js [-n] [-b] <file...>");
+  process.exit(1);
+}
+const showAllLineNumbers = args.includes("-n");
+const showNonBlankNumbers = args.includes("-b");
```
illicitonion, line 11: Imagine if I called this programme with a different flag, e.g. `node cat.js -x /some/file`. I probably passed that `-x` flag expecting it would do something. But your programme just ignores it.

What do you think we should do if we run into a flag our programme doesn't understand?

This applies to all three tools.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,54 @@
...
+
+const filePaths = args.filter((arg) => !arg.startsWith("-"));
+
+let lineNumber = 1;
+
+for (const filePath of filePaths) {
+  const content = await fs.readFile(filePath, "utf-8");
+
+  if (!showAllLineNumbers && !showNonBlankNumbers) {
```
illicitonion, line 29: These are really clear variable names, which makes this if statement read very clearly - well done!

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,54 @@
...
+
+    for (let i = 0; i < lines.length; i++) {
+      const line = lines[i];
+
+      let isLastLine;
+      if (i === lines.length - 1) {
+        isLastLine = true;
+      } else {
+        isLastLine = false;
+      }
```
illicitonion: This could be a bit more concise - generally every time you have code of the form:
```js
if (condition) {
  true
} else {
  false
}
```

you can probably use the fact that `condition` is a boolean, to just use that expression. e.g.
```js
const isLastLine = i === lines.length - 1;
```

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,54 @@
...
+
+      if (isLastLine && line === "") {
+        break;
+      }
+
+      if (showAllLineNumbers) {
```
illicitonion: The code in these two branches is quite repetitive - both do padding, hard-code what number we're padding, etc.

Can you think how to write this in such a way that we don't need to repeat e.g. the padding twice?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,30 @@
+import { promises as fs } from "node:fs";
+import process from "node:process";
+
+const args = process.argv.slice(2);
+
+const showHidden = args.includes("-a");
+
+const paths = args.filter((arg) => !arg.startsWith("-"));
+
+const targetPath = paths.length > 0 ? paths[0] : process.cwd();
```
illicitonion: If I call this programme with more than one argument (e.g. `ls /tmp /dev`), it will ignore the second one and just list the first. As a user, what would you expect to happen here?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+if (showHidden) {
+  entries = [".", "..", ...entries];
+}
+
+entries.sort((a, b) => {
+  const cleanA = a.replace(/^\.+/, "");
+  const cleanB = b.replace(/^\.+/, "");
+  return cleanA.localeCompare(cleanB);
```
illicitonion: What's this "cleaning" doing? What would break if you removed it?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,82 @@
...
+for (const { filePath, lines, words, bytes } of results) {
+  if (noSpecificFlag) {
+    process.stdout.write(formatLine([lines, words, bytes], filePath) + "\n");
+  } else if (showLines) {
+    process.stdout.write(formatLine([lines], filePath) + "\n");
+  } else if (showWords) {
+    process.stdout.write(formatLine([words], filePath) + "\n");
+  } else if (showBytes) {
+    process.stdout.write(formatLine([bytes], filePath) + "\n");
+  }
```
illicitonion: This is very repetitive - can you think how to make this less repetitive?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,82 @@
...
+  const parts = counts.map((n) => String(n).padStart(width, " "));
+  return parts.join("") + " " + label;
+}
+
+for (const { filePath, lines, words, bytes } of results) {
+  if (noSpecificFlag) {
+    process.stdout.write(formatLine([lines, words, bytes], filePath) + "\n");
+  } else if (showLines) {
+    process.stdout.write(formatLine([lines], filePath) + "\n");
+  } else if (showWords) {
```
illicitonion: What happens with the real `wc` if I pass multiple flags (e.g. `wc -wc /some/file`)? What does your implementation do?

illicitonion (commented):
Generally looking good, but some bits of duplication to think about :)

implement-shell-tools/cat/cat.js
Konvaly: Thanks for pointing this out! You're right, silently ignoring unknown flags is confusing for the user. I've added a check in all three tools that shows an error message and exits if the user passes a flag we don't recognise. For example, node cat.js -x file.txt will now print cat: invalid option -- 'x' instead of just running as if nothing happened.

implement-shell-tools/cat/cat.js
Konvaly: Thank you!

implement-shell-tools/cat/cat.js
Konvaly: Ah yes, that makes total sense! I've simplified it. Much cleaner, thanks for the tip!

implement-shell-tools/cat/cat.js
Konvaly: You're right,  I've refactored it. Now I first figure out whether the line needs a number using const needsLineNumber = showAllLineNumbers || (showNonBlankNumbers && !isBlankLine), and then do the padding only once inside a single if block. This way if I ever need to change the formatting, I only change it in one place.

implement-shell-tools/ls/ls.js
Konvaly: I'd definitely expect it to show both directories. I've fixed it , now it loops through all the paths, not just the first one. It also shows directory names as headers when there's more than one, like the real ls does.

implement-shell-tools/ls/ls.js
Konvaly: This removes leading dots from filenames before comparing them in the sort. Without it   .hidden.txt would appear before 1.txt because . comes before numbers in ASCII.

implement-shell-tools/wc/wc.js
Konvaly: Agreed. I pulled the logic into a getCounts() function that builds the right array depending on which flags are set. So now instead of four near-identical branches, it's just two lines.

implement-shell-tools/wc/wc.js
Konvaly: Real wc -wc shows both words and bytes. Mine didn't because the else if meant only one could win. Fixed it, combined flags like -wc and -lw work now.

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,56 @@
...
+}
+const showAllLineNumbers = args.includes("-n");
+const showNonBlankNumbers = args.includes("-b");
+const supportedFlags = ["-n", "-b"];
+const unknownFlags = args.filter(
+  (arg) => arg.startsWith("-") && !supportedFlags.includes(arg),
+);
+
+if (unknownFlags.length > 0) {
+  console.error(`cat: invalid option -- '${unknownFlags[0].slice(1)}'`);
```
illicitonion, line 18: Optional: It's often useful to tell people about _all_ of the problems you find, if it's easy to do so, so they can fix all of them - the experience of running:

```console
% cat -f -x /tmp/t
cat: invalid option -- f
% cat -x /tmp/t
cat: invalid option: -- x
% cat /tmp/t
It worked!
```

is quite frustrating compared to:

```console
% cat -f -x /tmp/t
cat: invalid options: f, x
% cat /tmp/t
It worked!
```

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,59 @@
...
+  if (targetPaths.length > 1) {
+    process.stdout.write(`${targetPath}:\n`);
+  }
+
+  let entries;
+
+  try {
+    entries = await fs.readdir(targetPath);
+  } catch (error) {
+    console.error(
```
illicitonion, line 32: Currently if there's an error (e.g. if you specify a path that doesn't exist), the exit code of this program will be 0. Is that what we want?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,85 @@
...
+
+const supportedFlags = ["-l", "-w", "-c"];
+const unknownFlags = expandedArgs.filter(
+  (arg) => arg.startsWith("-") && !supportedFlags.includes(arg),
+);
+
+if (unknownFlags.length > 0) {
+  console.error(`wc: invalid option -- '${unknownFlags[0].slice(1)}'`);
+  process.exit(1);
+}
```
illicitonion, line 29: This is good code, and works well considering different edge cases (well done!), but is a lot of code you've had to write to handle this :) you don't need to for this exercise, but you may want to look at a library like https://github.com/tj/commander.js which can do a lot of this for you.

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,85 @@
...
+
+const totalLines = results.reduce((sum, r) => sum + r.lines, 0);
+const totalWords = results.reduce((sum, r) => sum + r.words, 0);
+const totalBytes = results.reduce((sum, r) => sum + r.bytes, 0);
+
+function getCounts(lines, words, bytes) {
+  const counts = [];
+  if (noSpecificFlag || showLines) counts.push(lines);
+  if (noSpecificFlag || showWords) counts.push(words);
+  if (noSpecificFlag || showBytes) counts.push(bytes);
```
illicitonion, line 64: These work, but quite tightly couple the CLI flags to the behaviour here. We don't need to use only what flags we happened to expose to users when writing our logic - we can transform things ourselves. Can you think how you could set up `showLines`, `showWords` and `showBytes` when parsing the flags so that these checks would just read:

```js
if (showLines) counts.push(lines);
if (showWords) counts.push(words);
if (showBytes) counts.push(bytes);
```

but still work correctly when not flags were passed?

illicitonion (commented):
This is looking great, well done!

A couple of small things to pick up, and then a few things to think about that don't block marking this as complete :)

PR #502, opened 2026-04-18
London | 26-SDC-March | Zobeir Rigi | Sprint 3 | implement-shell-tools

Zobeir-Rigi:
Self checklist

Shell tools implementation: cat, ls, and wc

### DIFF
--- a/implement-shell-tools/ls/ls.js
+++ b/implement-shell-tools/ls/ls.js
@@ -3,14 +3,27 @@ import { promises as fs } from "node:fs";
 
 const args = process.argv.slice(2);
 
+let showAll = false;
 let path = ".";
 
-if (args[1]) {
-  path = args[1];
+for (const arg of args) {
+
+  if (arg === "-a") {
+    showAll = true;
+  }
+
+  else if (arg !== "-1") {
+    path = arg;
+  }
 }
 
 const files = await fs.readdir(path);
 
 for (const file of files) {
+
+  if (!showAll && file.startsWith(".")) {
+    continue;
+  }
+
   console.log(file);
 }
--- a/implement-shell-tools/ls/ls.js
+++ b/implement-shell-tools/ls/ls.js
@@ -1,0 +1,16 @@
+import process from "node:process";
+import { promises as fs } from "node:fs";
+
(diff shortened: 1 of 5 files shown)


implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,29 @@
...
+
+const files = await fs.readdir(path);
+
+for (const file of files) {
+
+  if (!showAll && file.startsWith(".")) {
+    continue;
+  }
+
+  console.log(file);
```
LonMcGregor: Are you correctly handling the -1 argument?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const lines = content.split("\n").length;
+  const words = content.split(" ").length;
+  const chars = content.length;
+
+  if (option === "l") {
+    console.log(lines, path);
+  } else if (option === "w") {
+    console.log(words, path);
+  } else if (option === "c") {
```
LonMcGregor: You are using else if here. What if someone wanted words and chars but not lines?

LonMcGregor (commented):
Can you try rewriting these using a package to manage the arguments? This can make it easier to handle options and command line access.

implement-shell-tools/ls/ls.js
Zobeir-Rigi: I’ve fixed it. -1 is handled now. When it’s used, nothing changes since console.log already prints one per line.

implement-shell-tools/wc/wc.js
Zobeir-Rigi: Good point, I was only handling one option. I’ve updated it so multiple flags like -w -c work now.

LonMcGregor (commented):
Think about why we have the -1 argument. What do we do if we want to change whether it prints on one line or over individual lines?

About WC, you now offer different argument combinations which is good. You might consider using a library/module to handle arguments as this makes it a lot easier.

LonMcGregor (commented):
Great work. I hope it is clear why using something like commander in this case is helpful.

illicitonion:

PR #503, opened 2026-04-18
London | 26-SDC-Mar | Khilola Rustamova| Sprint 4 |Implement cowsay

HilolaRustam:
## Learners, PR Template

Self checklist

## Changelist

Implement cowsay exercise is completed.


implement-cowsay/requirements.txt
LonMcGregor, line 1: Is this file missing something?

LonMcGregor (commented):
Part of your implementation may be missing, can you check?

implement-cowsay/requirements.txt
LonMcGregor, line 1: It looks like this file is still empty. If you're using a 3rd party package,it should be listed here

implement-cowsay/cow.py
```diff
@@ -0,0 +1,21 @@
...
+    choices=cowsay.char_names,
+    help="The animal to be saying things.")
+
+args = parser.parse_args()
+
+message = " ".join(args.message)
+
+animal = args.animal or "cow"
+
+print(getattr(cowsay, animal)(message))
```
LonMcGregor: Do you need this print() here?

LonMcGregor (commented):
Almost there! Good work

implement-cowsay/cow.py
HilolaRustam: Yes, cowsay returns a string rather than printing directly, so print() is needed. I’ll refactor it slightly for readability by separating the function call from the print statement.

implement-cowsay/requirements.txt
HilolaRustam: Thanks, I've added cowsay to requirements.txt so the dependency is properly declared.

LonMcGregor (commented):
Great,there's just one thing left. When you run this, do you see the final line says "None"? Why does it print this?

LonMcGregor (commented):
Great! the function prints the output by itself.

illicitonion:

PR #509, opened 2026-04-22
West Midlands | 26 March SDC | Iswat Bello | Sprint 5 | prep exercises for Object-Oriented Programming and Type safety

Iswanna:
## Learners, PR Template

Self checklist

## Changelist
Complete implementation of 12 preparatory exercises covering Python fundamentals, including classes, inheritance, type hints, generics, and enums.

### What's New
- 12 new Python exercise files demonstrating OOP concepts
- Type-safe implementations using dataclasses and enums
- Interactive user input examples

### Files Changed
- Added 11 `.py` exercise files
- Added 1 `.md` documentation file

### All Tests Pass ✓
- mypy type checking: PASS
- Runtime execution: PASS


prep/enums-exercise.py
behnamkvl, line 1: - The requemnets had this part "should output errors to stderr and terminate the program with a non-zero exit code if the user input bad values" but in the code it just returns and does not exit with non-zero code.
- If user puts an invalid input for the age the application crashes, think about how you can handle that better.

prep/generics-exercise1.py
behnamkvl, line 1: It is a clean fix, however, because the requrement was "do not change line 17", try to see what you can do for that, we want to access age like a property and not method!
Tip: Have a search on @property decorator usage.

behnamkvl (changes requested):

prep/enums-exercise.py
Iswanna: I have updated the code to address your feedback. I added a try/except block to handle the ValueError for the age input to prevent the application from crashing on non-numeric input. I also imported sys so that I could use sys.stderr for error messages and sys.exit(1) to ensure the program terminates with a non-zero exit code as requested. 
Thanks for the feedback.

prep/generics-exercise1.py
Iswanna: I have updated the code to satisfy the requirement of not changing Line 17. By adding the @property decorator to the age method, I've enabled the calculation to be accessed as a property `(child.age)` rather than a method call `(child.age()`.

Thank you @behnamkvl for the feedback.

behnamkvl (approved):
Good job on making the changes. The current implementation looks good to me.

Iswanna:
> Good job on making the changes. The current implementation looks good to me.

Thank you so much for the feedback.

behnamkvl (approved):

illicitonion:

PR #518, opened 2026-05-05
London | 26-Mar-SDC | Onur Atas | Sprint 3| Implement shell tools

onurat:
Self checklist

- [x] My changes follow the [style guide]

Implemented ls, cat, and wc in Node.js by manually parsing CLI arguments and replicating core Unix command behaviour.

### DIFF
--- a/implement-shell-tools/ls/ls.js
+++ b/implement-shell-tools/ls/ls.js
@@ -1,0 +1,25 @@
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
+const showAll = args.includes("-a");
+
+const dirArg = args.find(arg => !arg.startsWith("-"));
+const directory = dirArg || ".";
+
+fs.readdir(directory, (err, files) => {
+  if (err) {
+    console.error(err.message);
+    return;
+  }
+
+  let output = files;
+
+  if (!showAll) {
+    output = files.filter(file => !file.startsWith("."));
+  } else {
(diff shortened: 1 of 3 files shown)


implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,25 @@
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
+const showAll = args.includes("-a");
```
LonMcGregor, line 5: This task also wants you to handle the -1 argument. Can you try that?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,58 @@
...
+  const content = fs.readFileSync(file, "utf8");
+  const stats = getStats(content);
+
+  totalLines += stats.lines;
+  totalWords += stats.words;
+  totalBytes += stats.bytes;
+
+  let output = [];
+
+  if (showLines) output.push(format(stats.lines));
```
LonMcGregor, line 37: This pattern of code is repeated, is there any way you can reduce that?

LonMcGregor (commented):
Good start on these tasks

Across your solutions you manually check for options in the input arguments. Do you know of any library based solutions that could help build program argument handling that is a bit easier to manage? It would also let you add help text, and combine single letter arguments together like you can on the command line.

See also my other remarks.

PR #520, opened 2026-05-12
Make test-sdc script more portable

illicitonion:
* Add a shebang because `-v` is a bash-ism not present in whatever my default shell is on macOS.
* Remove `--quiet` from `cmp` - macOS's `cmp` doesn't support it.


LonMcGregor (approved):
Good catch - if the flag isnt present on mac we should remove it

PR #521, opened 2026-05-12
Add automated tests for cowsay implementation

illicitonion:


LonMcGregor (commented):
Looks good, I think we should also do something like a grep for text that shouldn't be in there, like for "dragon", and fail if present,  to make sure they're pulling the choices in dynamically.

illicitonion:
Good call - added that check (and a check that `--help` includes `dragon`)

LonMcGregor (approved):
Looks good

PR #524, opened 2026-05-19
London | 26-SDC-March | Zobeir Rigi | Sprint 4 | Implement shell tools python

Zobeir-Rigi:
Self checklist

### DIFF
--- a/implement-shell-tools/ls/ls.py
+++ b/implement-shell-tools/ls/ls.py
@@ -3,14 +3,22 @@ import os
 
 args = sys.argv[1:]
 
-# default folder = current folder
+show_all = False
 path = "."
 
-# if user gives a folder
-if len(args) > 1:
-    path = args[1]
+for arg in args:
+
+    if arg == "-a":
+        show_all = True
+
+    elif arg != "-1":
+        path = arg
 
 files = os.listdir(path)
 
 for file in files:
+
+    if not show_all and file.startswith("."):
+        continue
+
     print(file)
--- a/implement-shell-tools/ls/ls.py
+++ b/implement-shell-tools/ls/ls.py
@@ -1,0 +1,16 @@
+import sys
+import os
+
+args = sys.argv[1:]
+
+# default folder = current folder
+path = "."
+
+# if user gives a folder
+if len(args) > 1:
+    path = args[1]
+
+files = os.listdir(path)
+
+for file in files:
+    print(file)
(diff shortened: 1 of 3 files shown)


implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,24 @@
...
+
+for arg in args:
+
+    if arg == "-a":
+        show_all = True
+
+    elif arg != "-1":
+        path = arg
+
+files = os.listdir(path)
```
LonMcGregor: Part of this task is to see if you can handle the listing yourself.

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,24 @@
...
+
+show_all = False
+path = "."
+
+for arg in args:
+
+    if arg == "-a":
+        show_all = True
+
+    elif arg != "-1":
```
LonMcGregor: Is there any difference between when -1 is used and when not?

LonMcGregor (commented):

Is there a package you could use to help with managing arguments? Right now it is a bit complex, and you can't combine single-letter arguments like the command line tools.

implement-shell-tools/ls/ls.py
Zobeir-Rigi: Thank you, there wasn’t any difference before. I’ve now handled -1 explicitly.

implement-shell-tools/ls/ls.py
Zobeir-Rigi: Got it, added sorting to improve the listing.

LonMcGregor:
See my comments for the node.js version of this task - consider how to fully implement -1 and maybe use an argument parsing library

Zobeir-Rigi:
> Good start, but there is some work left to do here.
> 
> Is there a package you could use to help with managing arguments? Right now it is a bit complex, and you can't combine single-letter arguments like the command line tools.

Thanks — I’ve updated it to use argparse, which simplifies argument handling and supports combined flags

LonMcGregor (commented):
Good work. The main parts of this task are implemented and complete. If you want a stretch goal, you can take some extra time to conver the other parts to use argparse.

Zobeir-Rigi:
> Good work. The main parts of this task are implemented and complete. If you want a stretch goal, you can take some extra time to conver the other parts to use argparse.

Great, I will do it as soon as I get some time.

illicitonion:

PR #531, opened 2026-05-29
London | 26-Mar-SDC | Onur Atas | Sprint 4| implement shell tools

onurat:
Self checklist

## Changelist

Implemented Python versions of ls, cat, and wc, including the required flags and matching the expected shell command behaviour.


LonMcGregor (commented):

Is there a package you could use to help with managing arguments? Right now it doesn't offer any help text, and you can't combine single-letter arguments like the command line tools.

PR #532, opened 2026-05-29
London | 26-Mar-SDC | Onur Atas | Sprint 5| prep exercises

onurat:
Self checklist

## Changelist

Completed all Sprint 5 prep exercises:

* Why we use types
* Type checking with mypy
* Classes and objects
* Methods
* Dataclasses
* Generics
* Type-guided refactorings
* Enums
* Inheritance

Tested all exercises locally.

Note: `person_class.py` intentionally contains an invalid property access (`person.address`) because the exercise required demonstrating that mypy can detect access to a property that does not exist.

### DIFF
--- a/prep-exercises/enums.py
+++ b/prep-exercises/enums.py
@@ -1,0 +1,91 @@
+from dataclasses import dataclass
+from enum import Enum
+import sys
+
+
+class OperatingSystem(Enum):
+    MACOS = "macOS"
+    ARCH = "Arch Linux"
+    UBUNTU = "Ubuntu"
+
+
+@dataclass(frozen=True)
+class Person:
+    name: str
+    age: int
+    preferred_operating_system: OperatingSystem
+
+
+@dataclass(frozen=True)
+class Laptop:
+    id: int
+    manufacturer: str
+    model: str
+    operating_system: OperatingSystem
+
+
+laptops = [
+    Laptop(1, "Dell", "XPS", OperatingSystem.ARCH),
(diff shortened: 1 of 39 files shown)


LonMcGregor (commented):
I think you may have submitted the wrong files here, do you want to take another look?

onurat:
I made a mistake on the branching. This has all that's been done so far for the tools module. I believe only the prep exercises folder needs reviewing. Or you can look at this commit sha 0b59f61

prep-exercises/methods.py
```diff
@@ -0,0 +1,35 @@
+from datetime import date
+
```
LonMcGregor, line 2: Part of the methods task included giving an explanation of the differences between methods and functions. Can you give a brief explanation?

prep-exercises/enums.py
```diff
@@ -0,0 +1,91 @@
...
+]
+
+
+name = input("Name: ")
+
+try:
+    age = int(input("Age: "))
+except ValueError:
+    print("Invalid age", file=sys.stderr)
+    sys.exit(1)
```
LonMcGregor, line 41: Instead of exiting immediately, is there a way you could re-ask for correct input?

LonMcGregor (commented):
Good work on the prep exercises. 

I've left a couple of comments to guide you to a more complete solution.

It would be best if you can remove the unneeded files from this branch, otherwise we can't  accept this PR as complete.

PR #533, opened 2026-05-29
switching last two hex questions

Poonam-raj:
So a trainee has to convert the hex into a number before considering how this might look as colours.

Ideally to improve lead into the idea of that % of bytes used can create the colour mix.

Same could be said for the grey question needing more suggestion of how to calculate a colour mix (somehow hinting or leading to the idea of proportion of bytes used).

Suggestions welcome for whether this fixes a misunderstanding or not, or if it can be improved.


LonMcGregor (approved):
This makes sense

PR #536, opened 2026-06-08
London | 26-SDC-March | Zobeir Rigi | Sprint 5 | prep-exercises

Zobeir-Rigi:

sprint 5 - prep exercises, learned about type in Python, mypy, and OOP.

### DIFF
--- a/sprint-five-prep/enums.py
+++ b/sprint-five-prep/enums.py
@@ -52,20 +52,32 @@ def most_available_os() -> OperatingSystem:
 
 
 # Main program
-try:
-    name = input("Enter your name: ").strip()
 
-    age_input = input("Enter your age: ").strip()
-    age = int(age_input)
-    if age < 0:
-        raise ValueError("Age must be a positive number")
+name = input("Enter your name: ").strip()
 
-    os_input = input("Enter OS (macOS, Arch Linux, Ubuntu): ").strip()
-    preferred_os = parse_os(os_input)
+age = None
+preferred_os = None
 
-except ValueError as e:
-    print(f"Error: {e}", file=sys.stderr)
-    sys.exit(1)
+while age is None or preferred_os is None:
+    try:
+        if age is None:
+            age_input = input("Enter your age: ").strip()
+            age = int(age_input)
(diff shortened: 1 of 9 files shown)


sprint-five-prep/enums.py
```diff
@@ -0,0 +1,94 @@
...
+    age_input = input("Enter your age: ").strip()
+    age = int(age_input)
+    if age < 0:
+        raise ValueError("Age must be a positive number")
+
+    os_input = input("Enter OS (macOS, Arch Linux, Ubuntu): ").strip()
+    preferred_os = parse_os(os_input)
+
+except ValueError as e:
+    print(f"Error: {e}", file=sys.stderr)
```
LonMcGregor: If you get an input error, is there a user friendly way to handle this without exiting the program?

sprint-five-prep/methods.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+
+# ✍️exercise
+# Think of the advantages of using methods instead of free functions.
+#  Write them down in your notebook.
+# Encapsulation 
+# The class controls its own logic
+# Example: You change from age → date_of_birth
+# Only the class needs to change,  External code still works:person.is_adult()
```
LonMcGregor, line 43: Could you expand on this point, what do you mean by "external code still works"?

LonMcGregor (commented):
Good work on these tasks, just one part where code could be improved and another where explanation could be improved.

sprint-five-prep/enums.py
Zobeir-Rigi: > If you get an input error, is there a user friendly way to handle this without exiting the program?

I handled input errors by retrying in a loop instead of exiting, so users can fix mistakes without restarting.

sprint-five-prep/methods.py
Zobeir-Rigi: > Could you expand on this point, what do you mean by "external code still works"?

It means we can change the internal implementation inside the class to modify the method’s behaviour, without changing how external code uses it, for example person.is_adult() stays the same.

PR #537, opened 2026-06-10
London | 26-SDC-Mar | Khilola Rustamova| Sprint 4 |Implement shell tools python

HilolaRustam:
## Learners, PR Template

Self checklist

## Changelist

implementation of cat, ls and wc with python


implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,76 @@
...
+    for p in paths:
+        matches = glob.glob(p)
+        if matches:
+            files.extend(matches)
+        else: 
+            files.append(p) # keep as-is (will error later if missing)
+    return sorted(files)
+
+def read_lines(file):
+           with open(file, "r", encoding="utf-8") as f:
```
illicitonion: This block is more indented than the others - why is that?

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,76 @@
...
+               return f.readlines()
+       
+def print_lines(files, number_all=False, number_nonempty=False):
+    line_no = 1     
+    for file in files:
+        try:
+            lines = read_lines(file)
+        except FileNotFoundError:
+            print(f"cat: {file}: No such file or directory", file=sys.stderr)
```
illicitonion, line 45: Good job printing to stderr not stdout :)

However! If you ran into an issue here, I think your program will still exit with exit code 0. What exit code do you think it should exit with?

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,76 @@
...
+                prefix = f"{line_no:6}\t"
+                line_no += 1
+            else:
+                prefix = ""
+                
+            # avoid double newlines: line already includes '\n'
+            sys.stdout.write(prefix + line)
+            
+            
+def main():
```
illicitonion: This all works, but often people will have their main function at the _top_ of the file so it's the first thing you read, and allow you to scroll down to read details if needed. You don't _need_ to follow this, but it can often make it easier to read/review code.

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,76 @@
...
+        if a == "-n":
+            number_all = True
+        elif a == "-b":
+            number_nonempty = True
+            number_all = False #-b overrides -n
+        else:
+            paths.append(a)
+            
+    files = expand_paths(paths)
+    print_lines(files, number_all,number_nonempty)
```
illicitonion: I've noticed your whitespace is a bit inconsistent (e.g. no space after this `,`) - you may want investigate a tool like [`black`](https://github.com/psf/black) to help automate fixing this.

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,76 @@
...
+    number_all = False
+    number_nonempty = False
+    paths = []
+    
+    for a in args:
+        if a == "-n":
+            number_all = True
+        elif a == "-b":
+            number_nonempty = True
+            number_all = False #-b overrides -n
```
illicitonion: What would happen if `-b` appeared before `-n` in the command line?

implement-shell-tools/cat/mycat.py
illicitonion: We often used named arguments in Python when we have multiple arguments of the same type, to make sure we don't accidentally pass them in the wrong order. If you wrote `print_lines(files, number_nonempty, number_all)` here it wouldn't be obvious that the order was swapped. Instead we can write:

```python
print_lines(files, number_all=number_all, number_nonempty=number_nonempty)
```

and if we swapped them by mistake it would be obvious.

implement-shell-tools/ls/my-ls.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+
+def list_dir(path, show_all=False, one_per_line = False):
+    try:
+        entries = os.listdir(path)
+    except FileNotFoundError:
```
illicitonion: Same question as above about exit codes.

illicitonion (commented):
This is generally looking good, and like things work :) I left a few style comments, and many of my comments to one project apply to the others as well.

implement-shell-tools/cat/mycat.py
HilolaRustam: That makes sense. I've updated the call to use named arguments

HilolaRustam: Thank you for your suggestion. I've cleaned up the whitespace issues and I'll take a look at Black for automatic formatting

implement-shell-tools/cat/mycat.py
HilolaRustam: I've changed the flag handling so -b always takes precedence regardless of argument order.

implement-shell-tools/cat/mycat.py
HilolaRustam: Thank you, I've moved main() higher in the file to make the program flow easier to see.

implement-shell-tools/cat/mycat.py
HilolaRustam: You're right. A missing file should result in a non-zero exit code to indicate an error. I've updated the code to exit with code 1 if any file cannot be read.

implement-shell-tools/cat/mycat.py
HilolaRustam: That was an accident, i have aligned it.

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,89 @@
...
+    for a in args:
+        if a == "-n":
+            number_all = True
+        elif a == "-b":
+            number_nonempty = True
+        else:
+            paths.append(a)
+            
+    if number_nonempty:
+        number_all = False
```
illicitonion: This kind of relationship where different options/states are related but exclusive can be hard to follow. (Here, it should never be the case that `number_all` and `number_nonempty` are both `True` - it's an invalid state in the program).

Instead of this, we sometimes use enums for this (or can use strings as enums). Consider a single variable `number_mode` which could be set to either `"none"`, `"non_empty"` or `"all"` - here we don't need to think about what both `True` mean - we just have one variable which could have one of three variables. What do you think about this?

implement-shell-tools/wc/my-wc.py
```diff
@@ -0,0 +1,104 @@
...
+
+    byte_count = len(content)
+    text = content.decode("utf-8", errors="ignore")
+
+    line_count = text.count("\n")
+    word_count = len(text.split())
+
+    return line_count, word_count, byte_count
+
+def expand(paths):
```
illicitonion: In general you shouldn't need to do this kind of glob expansion yourself - the shell should do this for you and expand the globs into different file names when calling your program. You can test this out by printing what `sys.argv` is when you run `python3 my-wc.py some*glob` - look at what arguments you actually get :)

implement-shell-tools/wc/my-wc.py
```diff
@@ -0,0 +1,104 @@
...
+    total_w = 0
+    total_c = 0
+    
+    results = []
+    
+    for file in files:
+        try:
+            l, w, c = count_file(file)
+        except FileNotFoundError:
+            print(f"wc: {file}: No such file or directory", file=sys.stderr)
```
illicitonion: What exit code will you exit with if you failed to read one file?

illicitonion (commented):
This looks really good, well done!

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,89 @@
...
+        files, 
+        number_all=number_all,
+        number_nonempty=number_nonempty,
+        )
+
+    if had_error:
+        sys.exit(1)
+
+    
+def expand_paths(paths):
```
illicitonion: You shouldn't need to do this expansion in your program; the shell should do it for you.

implement-shell-tools/cat/mycat.py
illicitonion: I don't think you addressed this comment?

illicitonion (commented):
This is generally looking really good - just a couple of last things :)

implement-shell-tools/cat/mycat.py
```diff
@@ -0,0 +1,72 @@
...
+            
+            if number_mode == "non_empty":
+                if not is_empty:
+                    prefix = f"{line_no:6}\t"
+                    line_no += 1
+                else:
+                    prefix = "" 
+            elif number_mode=="all":
+                prefix = f"{line_no:6}\t"
+                line_no += 1
```
illicitonion, line 60: These branches look very similar - could they be combined?

implement-shell-tools/wc/my-wc.py
```diff
@@ -0,0 +1,104 @@
...
+    # print total if multiple files
+    if len(results) > 1:
+        parts = []
+
+        if show_l:
+            parts.append(f"{total_l:3}")
+        if show_w:
+            parts.append(f"{total_w:4}")
+        if show_c:
+            parts.append(f"{total_c:4}")
```
illicitonion, line 94: These blocks feel quite repetitive. Can you work out how to deduplicate them?

One thing that may be useful to know is you can define _nested_ functions which can access any variables within the scope they're defined in. e.g. you can write:

```python
def outer():
  value = 1

  def inner(name):
    print(f"{name}: {value}")

  for letter in ["a", "b", "c"]:
    inner(letter)
```

and `inner` can access `value`.

illicitonion (commented):
Sorry for the _huge_ delay here. This is looking really good - a couple of last things!

PR #540, opened 2026-06-15
London | 26-March-SDC | Ammad Ur Rehman | Sprint 4 | Implement shell tools (cat, ls, wc) in Python

anosidium:
## Learners, PR Template

Self checklist

## Changelist

Reimplement the shell programs as a Python program.

### DIFF
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -41,8 +41,12 @@ for path in args.paths:
         print(f"cat: {path}: Is a directory")
         continue
 
-    with open(path, "r") as file:
-        line_number = 1
+    try:
+        with open(path, "r") as file:
+            line_number = 1
         
-        for line in file:
-            line_number = process_line(line_number, line)
+            for line in file:
+                line_number = process_line(line_number, line)
+
+    except FileNotFoundError:
+        print(f"cat: {path}: No such file or directory")
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 
 parser = argparse.ArgumentParser(prog="cat",
                                  description="simple cat clone")
@@ -18,3 +19,30 @@ parser.add_argument("-b",
 parser.add_argument("paths", nargs="+", help="file(s) to process")
 
 args = parser.parse_args()
+
+def process_line(line_number, line):
+    if args.number_nonblank:
+        if line == "\n":
+            print(line, end="")
+            return line_number
+        
+        print(f"\t{line_number}  {line}", end="")
+        return line_number + 1
+    
+    if args.number_all:
+        print(f"\t{line_number}  {line}", end="")
+        return line_number + 1
+    
+    print(line, end="")
+    return line_number
+
+for path in args.paths:
+    if os.path.isdir(path):
+        print(f"cat: {path}: Is a directory")
+        continue
+
+    with open(path, "r") as file:
+        line_number = 1
+        
+        for line in file:
+            line_number = process_line(line_number, line)
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -26,11 +26,11 @@ def process_line(line_number, line):
             print(line, end="")
             return line_number
         
-        print(f"\t{line_number}  {line}", end="")
+        print(f"{line_number:6}\t{line}", end="")
         return line_number + 1
     
     if args.number_all:
-        print(f"\t{line_number}  {line}", end="")
+        print(f"{line_number:6}\t{line}", end="")
         return line_number + 1
     
     print(line, end="")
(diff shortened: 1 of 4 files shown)


implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,48 @@
...
+    print(line, end="")
+    return line_number
+
+for path in args.paths:
+    if os.path.isdir(path):
+        print(f"cat: {path}: Is a directory")
+        continue
+
+    with open(path, "r") as file:
+        line_number = 1
```
SlideGauge: README requires cat -n sample-files/*.txt to behave like real cat -n, which numbers lines continuously across all files

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,48 @@
...
+
+args = parser.parse_args()
+
+def process_line(line_number, line):
+    if args.number_nonblank:
+        if line == "\n":
+            print(line, end="")
+            return line_number
+        
+        print(f"\t{line_number}  {line}", end="")
```
SlideGauge: Output format for numbered lines doesn't match real cat. Real cat -n prints the number right-justified in a
6-wide field, then a tab, then the line

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,48 @@
...
+    
+    print(line, end="")
+    return line_number
+
+for path in args.paths:
+    if os.path.isdir(path):
+        print(f"cat: {path}: Is a directory")
+        continue
+
+    with open(path, "r") as file:
```
SlideGauge: What happens if the file is absent? Compare it with real cat behavior

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,33 @@
...
+                    action="store_true",
+                    help="show hidden files")
+
+parser.add_argument("filepath",
+                    nargs="?",
+                    default=".")
+
+args = parser.parse_args()
+
+entries = os.listdir(args.filepath)
```
SlideGauge: Does os.listdir return dir list in an alphabetical order? Compare with what real ls does

implement-shell-tools/ls/ls.py
```diff
@@ -0,0 +1,33 @@
...
+
+if not args.show_hidden_files:
+    entries = [entry for entry in entries if not entry.startswith(".")]
```
SlideGauge, line 28: -a filters out names starting with ., but real ls -a also includes . and .. as the first two entries.

SlideGauge (commented):
Good job, left several comments, could you fix them please?

implement-shell-tools/cat/cat.py
anosidium: Real `cat` produces:

```bash
cat -n sample-files/*.txt

     1  Once upon a time...
     1  It looked delicious.
     4
```

Python `cat` produces:

```bash
python3 cat.py -n sample-files/*.txt

        1  Once upon a time...
        1  It looked delicious.
        4  
```

The only difference is the indentation.

implement-shell-tools/ls/ls.py
anosidium: When I run ls -a, it produces:

```bash
.               ls.py           README.md
..              node_modules    sample-files
```

When I run the Python variant, python3 ls.py -a:

```bash
.       ..      README.md       ls.py   node_modules    sample-files
```

The alphabetically ordering does not match.

anosidium:
Thanks for the review. I've made some changes.

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,52 @@
...
+    return line_number
+
+for path in args.paths:
+    if os.path.isdir(path):
+        print(f"cat: {path}: Is a directory")
+        continue
+
+    try:
+        with open(path, "r") as file:
+            line_number = 1
```
SlideGauge, line 46: numbering still restarts per file, not continuous.

anosidium: Sorry, I’m going to have to disagree with this one. The program matches the behaviour of `cat` on my system. I checked all the commands in `README.md` and the output matches.

If I move `line_number = 1` outside the loop, the numbering becomes continuous across files and no longer matches `cat`.

My Python implementation:

     1  Once upon a time...
     2  There was a house made of gingerbread.
     3  It looked delicious.
     4  I was tempted to take a bite of it.
     5  But this seemed like a bad idea...
     6
     7  There's more to come, though...

cat:

     1  Once upon a time...
     1  It looked delicious.
     4

So keeping `line_number = 1` inside the per-file loop is what makes my implementation match `cat` on macOS. This may be a GNU/Linux vs BSD/macOS difference.

anosidium:
@cjyuan Could you please review my pull request? Thank you.

anosidium:
@SlideGauge Could you please review my pull request? Thank you.

PR #567, opened 2026-07-10
Manchester | 26-SDC-Jul | Fithi-Teklom | Sprint 1 | Individual-shell-tools

Fithi-Teklom:
Self checklist

## Changelist

Completed the shell tools exercises covering ls, cat, wc, grep, sed, and awk.

What was added:

- Implemented commands to list files and directories using ls.
- Used cat to read and display file contents.
- Used wc to count lines, words, and characters.
- Used grep to search and filter text based on patterns.
- Used sed for text replacement, deletion, and formatting transformations.
- Used awk to extract columns, filter data, and perform calculations.

Task ID: CYF-1145


individual-shell-tools/cat/script-01.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the contents of the helper-1.txt file inside the helper-files directory to the terminal.
 # The output of this command should be "Once upon a time...".
+
+cat helper-files/helper-1.txt
```
SlideGauge: helper-files/ is a sibling of these folders, not a child.

individual-shell-tools/cat/script-03.sh
```diff
@@ -9,3 +9,5 @@ set -euo pipefail
 # 1 It looked delicious.
 # 2 I was tempted to take a bite of it.
 # 3 But this seemed like a bad idea...
+
+cat -n helper-files/helper-3.txt
```
SlideGauge: helper-files/ is a sibling of these folders, not a child.

individual-shell-tools/cat/script-02.sh
```diff
@@ -11,3 +11,5 @@ set -euo pipefail
 # It looked delicious.
 # I was tempted to take a bite of it.
 # But this seemed like a bad idea...
+
+cat helper-files/*
```
SlideGauge: helper-files/ is a sibling of these folders, not a child.

individual-shell-tools/wc/script-01.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the number of words in the file helper-files/helper-3.txt.
 # The output should include the number 19. The output should not include the number 92.
+
+wc -w helper-files/helper-3.txt
```
SlideGauge: helper-files/ is a sibling of these folders, not a child.

individual-shell-tools/wc/script-02.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the number of lines in the file helper-files/helper-3.txt.
 # The output should include the number 3. The output should not include the number 19.
+
+wc -l helper-files/helper-3.txt
```
SlideGauge: helper-files/ is a sibling of these folders, not a child.

individual-shell-tools/wc/script-03.sh
```diff
@@ -8,3 +8,5 @@ set -euo pipefail
 #        1       7      39 ../helper-files/helper-2.txt
 #        3      19      92 ../helper-files/helper-3.txt
 #        5      30     151 total
+
+wc helper-files/*
```
SlideGauge: helper-files/ is a sibling of these folders, not a child.

individual-shell-tools/grep/script-01.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt said by the Doctor.
 # The output should contain 6 lines.
+
+grep "Doctor" dialogue.txt
```
SlideGauge: Does it grabs _only_ lines said by a doctor, not something like "Hello, Doctor!"?

individual-shell-tools/grep/script-07.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output, for each `.txt` file in this directory, how many lines of dialogue the Doctor has.
 # The output should show that dialogue.txt contains 6 lines, dialogue-2.txt contains 2, and dialogue-3.txt contains 0.
+
+#grep -c "Doctor" *.txt
```
SlideGauge: The line is commented, what was intended here to happen?

individual-shell-tools/sed/script-04.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output input.txt replacing every occurrence of the string "We'll" with "We will".
 # The output should contain 11 lines.
+
+sed "We'll/We will/g"
```
SlideGauge: How does it know which file to look into?

individual-shell-tools/ls/script-04.sh
```diff
@@ -16,8 +16,11 @@ echo "First exercise (sorted newest to oldest):"
 # TODO: Write a command which lists the files in the child-directory directory, one per line, sorted so that the most recently modified file is first.
 # The output should be a list of names in this order, one per line: helper-3.txt, helper-1.txt, helper-2.txt.
 
+# ls -1t child-directory
```
SlideGauge: Why is this line commented?

SlideGauge (commented):
A lot of files do not have trailing new line, add it please as well

SlideGauge:
Could you address my comments please?

Fithi-Teklom:
> Could you address my comments please?

I have corrected the mistakes.

individual-shell-tools/awk/script-01.sh
```diff
@@ -4,3 +4,4 @@ set -euo pipefail
 
 # TODO: Write a command to output just the names of each player in `scores-table.txt`.
 # Your output should contain 6 lines, each with just one word on it.
+awk '{print $1}' individual-shell-tools/awk/scores-table.txt
```
SlideGauge: 🔴 This path only works when the script is run from the repository root. Per the README, scripts are run from their own directory — so this should just be `scores-table.txt`. The same applies to all of your awk, grep and sed scripts.

individual-shell-tools/cat/script-01.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the contents of the helper-1.txt file inside the helper-files directory to the terminal.
 # The output of this command should be "Once upon a time...".
+
+cat individual-shell-tools/helper-files/helper-1.txt
```
SlideGauge: 🔴 Closer than before, but this still assumes the script is run from the repository root. From inside the `cat/` folder, you reach `helper-files/` with `../helper-files/helper-1.txt` — that's what "sibling" means: go up one level with `..`, then into the folder. Same fix for the other `cat` and `wc` scripts.

individual-shell-tools/grep/script-01.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output every line in dialogue.txt said by the Doctor.
 # The output should contain 6 lines.
+
+grep "Doctor" individual-shell-tools/grep/dialogue.txt
```
SlideGauge: 🔴 This is the same issue I raised in my previous review: `grep "Doctor"` matches *any* line containing the word, so you get 8 lines — including `Patient: Hello Doctor` and the Spouse's line — while the task expects the 6 lines *said by* the Doctor. You already solved this in script-07 with `^Doctor:` — use the same idea here.

individual-shell-tools/grep/script-06.sh
```diff
@@ -4,3 +4,5 @@ set -euo pipefail
 
 # TODO: Write a command to output the name of every `.txt` file in this directory which contains a line of dialogue said by the Doctor.
 # The output should contain two filenames.
+
+grep -l "Doctor" individual-shell-tools/grep/*.txt
```
SlideGauge: 🟡 This happens to output the right two files, but a bit by luck — `"Doctor"` matches any mention of the word, not only lines the Doctor says. `^Doctor:` (like your script-07) expresses exactly what the task asks and won't break if a file merely mentions the Doctor.

individual-shell-tools/awk/script-07-stretch.sh
```diff
@@ -7,3 +7,4 @@ set -euo pipefail
 # TODO: Write a command to output just the names of each player along with the total of adding all of that player's scores.
 # Your output should contain 6 lines, each with one word and one number on it.
 # The first line should be "Ahmed 15". The second line should be "Basia 37"
+awk '{total=0; for(i=3;i<=NF;i++) total += $i; print $1, total}' individual-shell-tools/awk/scores-table.txt
```
SlideGauge: 🟢 Nice work on the stretch goal — looping over the fields with `for(i=3;i<=NF;i++)` is exactly the right approach here.

individual-shell-tools/sed/script-05.sh
```diff
@@ -6,3 +6,5 @@ set -euo pipefail
 # If a line starts with a number and a space, make the line instead end with a space and the number.
 # So line 6 which currently reads "37 Alisha" should instead read "Alisha 37".
 # The output should contain 11 lines.
+
+sed 's/^\([0-9][0-9]*\) \(.*\)$/\2 \1/' individual-shell-tools/sed/input.txt
```
SlideGauge: 🟢 Good use of capture groups to swap the number and the name — this is a tricky exercise and your regex is correct.

SlideGauge (changes requested):
Thanks for the update — the command logic is mostly correct now, and the awk/sed solutions show good understanding. However the tests still fail 24/28, and the root cause is the same for almost every script: **file paths**.

The README for this exercise says:

> Before running a script you should `cd` into the directory it's in. You should write all of your scripts assuming they're running inside the directory they're saved in.

Paths like `individual-shell-tools/awk/scores-table.txt` only work when the script is run from the repository root — the tests run each script from its own folder. So:

- in `awk` scripts: `scores-table.txt`
- in `grep` scripts: `dialogue.txt` / `*.txt`
- in `sed` scripts: `input.txt`
- in `cat`/`wc` scripts: `../helper-files/...` (this is what my earlier "sibling" comments meant — from inside `cat/`, go **up** one level with `..`, then into `helper-files/`)

Your `ls` scripts already do this correctly — that's exactly why they are the 4 that pass.

Also, most files are still missing the trailing newline I asked about in my previous review — please add it to all scripts.

Tip for checking your work before pushing: `cd` into each folder and run the script, e.g. `cd individual-shell-tools/awk` then `./script-01.sh` — if it errors or prints the wrong thing there, the test will fail too.

SlideGauge:
Good job, review complete.

PR #593, opened 2026-07-22
London |26-JUL-SDC |Shaghayeghfar| Sprint 3 |Implement  Shell  Tools

shaghayeghfar:
## Learners, PR Template

Self checklist

## Changelist

Implemented simplified versions of common Unix shell commands  using  JavaScript and Node.js

Task ID: CYF-1150

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -8,12 +8,16 @@ let startIndex = 0;
 let lineNumber = 1;
 
 // Check for flags
-if (args[0] === "-n") {
-  numberAllLines = true;
-  startIndex = 1;
-} else if (args[0] === "-b") {
-  numberNonBlankLines = true;
-  startIndex = 1;
+for (let i = 0; i < args.length; i++) {
+  if (args[i] === "-n") {
+    numberAllLines = true;
+    startIndex++;
+  } else if (args[i] === "-b") {
+    numberNonBlankLines = true;
+    startIndex++;
+  } else {
+    break;
+  }
 }
 
 // Read each file
@@ -23,7 +27,10 @@ for (let i = startIndex; i < args.length; i++) {
   try {
     let content = fs.readFileSync(fileName, "utf8");
 
-    let lines = content.split("\n");
+    // Remove the extra empty line if the file ends with "\n"
+    let lines = content.endsWith("\n")
+      ? content.slice(0, -1).split("\n")
+      : content.split("\n");
 
     for (let j = 0; j < lines.length; j++) {
       let line = lines[j];
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,0 +1,48 @@
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
+let numberAllLines = false;
+let numberNonBlankLines = false;
+let startIndex = 0;
+let lineNumber = 1;
+
+// Check for flags
+if (args[0] === "-n") {
+  numberAllLines = true;
+  startIndex = 1;
+} else if (args[0] === "-b") {
+  numberNonBlankLines = true;
+  startIndex = 1;
+}
+
+// Read each file
+for (let i = startIndex; i < args.length; i++) {
+  let fileName = args[i];
+
+  try {
+    let content = fs.readFileSync(fileName, "utf8");
+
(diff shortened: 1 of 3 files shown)


implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,49 @@
...
+let path = "."; // Current directory by default
+
+// Check for flags
+for (let i = 0; i < args.length; i++) {
+  if (args[i] === "-1") {
+    onePerLine = true;
+  } else if (args[i] === "-a") {
+    showHidden = true;
+  } else {
+    path = args[i];
```
SlideGauge: When you run ls sample-files/*, the shell passes many paths as separate arguments. Each loop does path = args[i] - after the loop, how many of those paths are left? What does the real ls show for a glob like that, and how could you keep all of them rather than just one?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,68 @@
...
+
+  result += fileName;
+
+  console.log(result);
+}
+
+// Run wc for every file
+
+for (let file of files) {
+  countFile(file);
```
SlideGauge: Try the real wc sample-files/* with several files — what appears on the last line that your output doesn't have yet? Where in this loop could you accumulate the counts to produce that?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,68 @@
...
+// If no flags are given, wc shows all three
+if (!countLines && !countWords && !countBytes) {
+  countLines = true;
+  countWords = true;
+  countBytes = true;
+}
+
+// Function to count one file
+function countFile(fileName) {
+  const content = fs.readFileSync(fileName, "utf8");
```
SlideGauge: cat.js wraps its readFileSync in try/catch, but this one doesn't - and sample-files/* includes the dir subdirectory. What happens when readFileSync is handed a directory, and what does that do to the files listed after it?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Read each file
+for (let i = startIndex; i < args.length; i++) {
+  let fileName = args[i];
+
+  try {
+    let content = fs.readFileSync(fileName, "utf8");
+
+    let lines = content.split("\n");
```
SlideGauge: If a file ends in \n, what's the last element of content.split("\n")? Combined with console.log adding its own newline (line 42), how does your output compare to cat sample-files/1.txt - same number of lines, or one extra?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+try {
+  // Check if the path is a file
+  if (fs.statSync(path).isFile()) {
+    console.log(path);
+  } else {
+    let files = fs.readdirSync(path);
```
SlideGauge: readdirSync returns entries in whatever order the filesystem gives. Does the real ls do the same, or does it guarantee a particular order? What one call would make yours match?

implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,48 @@
...
+
+const args = process.argv.slice(2);
+
+let numberAllLines = false;
+let numberNonBlankLines = false;
+let startIndex = 0;
+let lineNumber = 1;
+
+// Check for flags
+if (args[0] === "-n") {
```
SlideGauge: Flags are only checked at args[0]. That covers the required command lines - but ask yourself: what would happen with -n file1 file2, or a flag that isn't first?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,68 @@
...
+
+  let lines = content.split("\n").length - 1;
+
+  let words = content
+    .trim()
```
SlideGauge: Nice — trim().split(/\s+/).filter(...) correctly returns 0 words for an empty file, an edge case that trips a lot of people up.

SlideGauge (commented):
Good job, I've got several notes from me, could you address them please?

shaghayeghfar:
Dear @SlideGauge 

Thank you very much for all your help, feedback, and the time you have taken to review my work. I really appreciate it. Thank you again for your support and guidance!

SlideGauge:
Good job, review complete!

PR #611, opened 2026-07-27
London | 26-SDC-July | Kris Oldrini | Sprint 3 | Implement shell tools

XiaoQuark:
## Learners, PR Template

Self checklist

## Changelist

- Completed `cat`, `ls` and `wc` implementation in JavaScript.

Task ID: CYF-1150


implement-shell-tools/cat/cat.mjs
```diff
@@ -0,0 +1,53 @@
+import process from "node:process";
+import { promises as fs } from "node:fs";
+import { Command } from "commander";
```
LonMcGregor, line 3: Is there a way you can get a program from commander directly?

implement-shell-tools/ls/ls.mjs
```diff
@@ -0,0 +1,126 @@
+import process from "node:process";
+import fs, { truncate } from "node:fs";
+
+const argv = process.argv.slice(2);
```
LonMcGregor, line 4: is there a reason you don't use the commander or parseargs api here?

LonMcGregor (commented):
Good work on these tasks, well organised code.I have some questions for you to push further

XiaoQuark:
Hi @LonMcGregor,
I have now amended the `cat` implementation and I'm importing program directly from commander.

Regarding the `ls` implementation, I wanted to experiment with handling flag parsing myself, so I intentionally chose not to use any libraries nor API.

LonMcGregor:
That makes sense, it's always good to practice. The task looks complete.

I would be interested if you could tell me some of the advantages or disadvantages you found when dealing with arguments in a nodejs app with and without a library?

PR #614, opened 2026-07-27
Birmingham | 26-JUL-SDC | Merve Reis | Sprint 3 | Sheel Tools

mervereis:
… features

## Learners, PR Template

Self checklist

Completed all the tasks.

Task ID: CYF-1150

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,0 +1,50 @@
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
+let numberLines = false;
+let numberNonBlank = false;
+const files = [];
+
+for (const arg of args) {
+  if (arg === "-n") {
+    numberLines = true;
+  } else if (arg === "-b") {
+    numberNonBlank = true;
+  } else {
+    files.push(arg);
+  }
+}
+
+if (numberNonBlank) {
+  numberLines = false;
+}
+
+let lineNumber = 1;
+
+for (const file of files) {
+  try {
+    const contents = fs.readFileSync(file, "utf8");
+    const lines = contents.split("\n");
+
+    lines.forEach((line, index) => {
+      const output = index < lines.length - 1 ? line + "\n" : line;
+
+      if (numberNonBlank) {
+        if (line.trim() === "") {
+          process.stdout.write(output);
+        } else {
+          process.stdout.write(`${String(lineNumber).padStart(6)}\t${output}`);
+          lineNumber++;
+        }
+      } else if (numberLines) {
+        process.stdout.write(`${String(lineNumber).padStart(6)}\t${output}`);
+        lineNumber++;
+      } else {
+        process.stdout.write(output);
+      }
+    });
+  } catch (err) {
+    console.error(`cat: ${file}: ${err.message}`);
+  }
+}
--- a/implement-shell-tools/ls/ls.js
+++ b/implement-shell-tools/ls/ls.js
@@ -1,0 +1,52 @@
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+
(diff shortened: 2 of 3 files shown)


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    const contents = fs.readFileSync(file, "utf8");
+    const lines = contents.split("\n");
+
+    lines.forEach((line, index) => {
+      const output = index < lines.length - 1 ? line + "\n" : line;
+
+      if (numberNonBlank) {
+        if (line.trim() === "") {
+          process.stdout.write(output);
```
LonMcGregor, line 35: When I am looking at the output  of blank lines I get some odd formatting, do you see this in your testing?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,52 @@
...
+          continue;
+        }
+        if (onePerLine) {
+          console.log(file);
+        } else {
+          process.stdout.write(file + " ");
+        }
+      }
+
+      if (!onePerLine) {
```
LonMcGregor, line 45: What is this last empty log doing?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+for (let file of files) {
+  countFile(file);
+}
+
+if (filesCounted > 1) {
+  let result = "";
+  if (countLines) {
```
LonMcGregor, line 70: This printing code looks to be repeated, do you think you could reduce the duplication somehow?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+  if (countWords) {
+    result += totalWords + " ";
+  }
+  if (countBytes) {
+    result += totalBytes + " ";
+  }
+  result += "total";
+
+  console.log(result);
```
LonMcGregor, line 81: When you have multiple files, is the output formatting as good as it could be? How could it be made neater?

LonMcGregor (commented):
Good start but there is still a lot to do here.

Is your way of handling options/arguments the best? Is there a better way of doing this?

implement-shell-tools/cat/cat.js
mervereis: Yes, I found the formatting issue in my testing. It was caused by the trailing newline being treated as an additional blank line. I fixed it so blank lines are handled correctly, and the output now matches the standard cat command. @LonMcGregor

implement-shell-tools/ls/ls.js
mervereis: It adds the final newline after the filenames have been written with process.stdout.write(), which does not add one automatically. The !onePerLine condition ensures this only happens in the default format, where filenames are displayed on the same line. @LonMcGregor

implement-shell-tools/wc/wc.js
mervereis: Yes, I reduced the duplication by moving the repeated output-building logic into a shared formatResult() function. Both the individual file results and the final total result now use this function, keeping the formatting consistent and easier to maintain. @LonMcGregor

implement-shell-tools/wc/wc.js
mervereis: The output could be neater. The numeric values should be right-aligned in consistent columns, matching the standard wc format. I updated the formatting so both individual file results and the total row use aligned columns, making the output easier to read. @LonMcGregor

LonMcGregor:
Thanks for responding with these changes and comments. This looks good now. You might consider trying the implementation again using a dedicated argument parsing library for practice, but this is enough for this task.

PR #620, opened 2026-07-28
Glasgow | 26-SDC-JUL | Mirabelle Morah | Sprint 3 | Implement Shell Tools

mirabellemorah:
## Learners, PR Template

Self checklist

## Changelist

Used node to implement shell command lines

Task ID: CYF-1150

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -2,66 +2,55 @@ import { program } from "commander";
 import { promises as fs } from "node:fs";
 
 program
-  .name("cat")
-  .description("Concatenate and print files")
+  .name("check-for-cat")
+  .description("Implement my own version of cat")
   .argument("<paths...>", "The file paths to process")
   .option("-n", "Number lines")
   .option("-b", "Number non-blank lines");
 
 program.parse();
 
-function parseNumberMode(options) {
-  if (options.b) {
-    return "non-blank";
-  } else if (options.n) {
-    return "all";
-  } else {
-    return "none";
-  }
-}
+const paths = program.args;
+const showN = program.opts().n;
+const showB = program.opts().b;
 
-function calculatePrefix(
-  numberMode,
-  nonBlankLineNumber,
-  lineNumberIncludingBlanks,
-  thisLineIsBlank,
-) {
-  if (numberMode === "none") {
-    return "";
-  }
-  let lineNumber;
-  if (numberMode === "all") {
-    lineNumber = lineNumberIncludingBlanks;
-  } else {
-    if (thisLineIsBlank) {
-      return "";
-    } else {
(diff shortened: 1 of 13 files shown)


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,67 @@
...
+for (const path of program.args) {
+  const content = await fs.readFile(path, "utf-8");
+  const lines = content.split("\n");
+  let nonBlankLineNumber = 1;
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    if (i === lines.length - 1 && line === "") {
+      break;
+    }
+    const prefix = calculatePrefix(
```
LonMcGregor: Is running calculatePrefix for every line of output an efficient solution?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
+import { promises as fs } from "node:fs";
+import process from "node:process";
+
+const argv = process.argv.slice(2);
```
LonMcGregor: Is there a reason you used commander in the cat example, but not here

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,47 @@
...
+
+for (const arg of argv) {
+  if (arg.startsWith("-")) {
+    flags.push(arg);
+  } else {
+    paths.push(arg);
+  }
+}
+
+const showOnePerLine = flags.includes("-1");
```
LonMcGregor: What does the showOnePerLine const do?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,70 @@
+import { program } from "commander";
+import { promises as fs } from "node:fs";
+
+const argv = process.argv.slice(2);
```
LonMcGregor: You import the commander program api, but don't use it. Is there a reason for that?

LonMcGregor (commented):

implement-shell-tools/cat/cat.js
mirabellemorah: No it wasn't, I reverted to the commander import

implement-shell-tools/ls/ls.js
mirabellemorah: Now showOneLine (line 13), when true it loops through all the files and prints them on a single line if -1 flag is passed (line 39-45), if not they'd print as it often does in multiple columns in vs code

implement-shell-tools/wc/wc.js
mirabellemorah: This has been corrected now to include arguments and options

LonMcGregor:
This looks good now, good work

PR #630, opened 2026-07-29
Glasgow | 26-SDC-July | Mansoor Munawar| Sprint 3 | Implement shell tools

MansoorM11:
## Self checklist

## Changelist

Implemented shell tools such as cat, ls and grep.

Task ID: CYF-1150


implement-shell-tools/ls/ls.mjs
```diff
@@ -0,0 +1,22 @@
...
+let files = await fs.readdir(targetDir);
+
+if (flags.includes("-a")) {
+  files.unshift(".", "..");
+} else {
+  files = files.filter((fileName) => !fileName.startsWith("."));
+}
+
+if (flags.includes("-1")) {
+  for (const file of files) {
```
LonMcGregor, line 17: is there a more advanced javascript way of printing each item in an array, so you don't have to write a full for loop?

implement-shell-tools/wc/wc.mjs
```diff
@@ -0,0 +1,48 @@
...
+  const content = await fs.readFile(filePath, "utf-8");
+
+  if (flag === "-w") {
+    console.log(getWordCount(content));
+  } else if (flag === "-l") {
+    console.log(getLineCount(content));
+  } else if (flag === "-c") {
+    console.log(getByteCount(content), filePaths);
+  } else {
+    console.log(
```
LonMcGregor, line 24: When I tested, i got output that looks very different to the original wc app. Can you find a way to make it look more similar?

LonMcGregor (commented):

You've built your own argument handling system. Is your way of handling options/arguments always going to be the best? Is there a better way of doing this?

PR #634, opened 2026-07-29
London | 26-JUL-SDC | Ping Wang | Sprint 4 | Implement-shell-tools

pathywang:

## Changelist

As usual, i did prep then made branch from main, create new file and did coursework with python language which is mirrored the same task with JavaScript language, still hard with AI explanatio and require lots of memorization and review

Task ID: CYF-1152

### DIFF
--- a/implement-shell-tools/wc/wc.py
+++ b/implement-shell-tools/wc/wc.py
@@ -1,0 +1,92 @@
+import sys
+
+args = sys.argv[1:]
+
+count_lines = False
+count_words = False
+count_bytes = False
+files = []
+
+for arg in args:
+    if arg == "-l":
+        count_lines = True
+    elif arg == "-w":
+        count_words = True
+    elif arg == "-c":
+        count_bytes = True
+    else:
+        files.append(arg)
+
+# If no flags are given, wc shows all three
+if not count_lines and not count_words and not count_bytes:
+    count_lines = True
+    count_words = True
+    count_bytes = True
+
+
+def count_file(filename):
+    try:
+        with open(filename, "r") as file:
+            content = file.read()
+
+        lines = content.count("\n")
+
+        words = 0 if content.strip() == "" else len(content.split())
+
+        # UTF-8 bytes, same idea as Buffer.byteLength()
+        bytes_count = len(content.encode("utf-8"))
+
+        output = []
+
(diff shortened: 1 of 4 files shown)


implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,92 @@
...
+
+        words = 0 if content.strip() == "" else len(content.split())
+
+        # UTF-8 bytes, same idea as Buffer.byteLength()
+        bytes_count = len(content.encode("utf-8"))
+
+        output = []
+
+        if count_lines:
+            output.append(str(lines))
```
LonMcGregor: If you run this with multiple files, how does the output look? What change could make it neater?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if counts:
+        total_lines += counts["lines"]
+        total_words += counts["words"]
+        total_bytes += counts["bytes"]
+
+
+if len(files) > 1:
+    output = []
+
```
LonMcGregor: What do you think about the output code here? Can you spot any duplication across the file?

LonMcGregor (commented):
Good start, but some things to fix.

also, Is the choice you have used for processing arguments the most effective? Is there a library you could use to help the implementation?

implement-shell-tools/wc/wc.py
pathywang: the colums do not line up very well because the numbers have different length so i formatted the string with fixted length  8.

implement-shell-tools/wc/wc.py
pathywang: yes there are nearly two identical codes so i define print_out function then call twice later on, refactor code

pathywang:
i manually use sys.agrv instead of import argparse  from librares, i noticed it then i would use import argparse later on which should be better. thanks

LonMcGregor (commented):
Good,the print function is a good way of reducing code duplication.

Yes, using the argparse library is more flexible - it also allows you to autocreate a help menu system. Keep it in mind for future tasks!

pathywang:
thank you  for explanation and i would do as your suggestion.

PR #635, opened 2026-07-30
London | 26-SDC-July | Alex Jamshidi | Sprint 3 | Implement shell tools 

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

cat
Handles all commands correctly:
cat sample-files/1.txt
cat -n sample-files/1.txt
cat sample-files/.txt
cat -n sample-files/.txt
cat -b sample-files/3.txt

Stretch:
Handles multiple flags and incorrect flags
(although does not parse b and n correctly when used together)

wc
Handles all commands correctly:
wc sample-files/*
wc -l sample-files/3.txt
wc -w sample-files/3.txt
wc -c sample-files/3.txt
wc -l sample-files/*
wc -w -l sample-files/3.txt
wc -w -l sample-files/*

Stretch:
Handles multiple flags and incorrect flags
Handles flag order

ls
Handles all commands correctly:
ls sample-files
ls sample-files/1.txt
ls sample-files/
ls -1
ls -1 sample-files
ls -1 -a sample-files

Stretch
Handles multiple flags and incorrect flags
Handles flag order

Task ID: CYF-1150

### DIFF
--- a/implement-shell-tools/cat/cat.js
+++ b/implement-shell-tools/cat/cat.js
@@ -1,46 +1,73 @@
 const fs = require("fs");
 
 const arguments = process.argv;
-const filePaths = arguments.slice(2);
+const userArguments = arguments.slice(2);
 
 const index = arguments[1].lastIndexOf("/");
 const currentWorkingDirectory = arguments[1].slice(0, index + 1);
 
-let flag = filePaths[0];
-if (flag.charAt(0) == "-") {
-  flag = filePaths.shift();
-}
+const flags = userArguments
+  .filter((argument) => argument.startsWith("-"))
+  .map((argument) => argument.slice(1))
+  .join("");
 
-function splitFile(fileName) {
-  const filePath = currentWorkingDirectory + fileName;
-  const text = fs.readFileSync(filePath, "utf8").trimEnd();
-  return text.split("\n");
-}
+const flagHandlers = {
+  b: bFlag,
+  n: nFlag,
+};
 
-function printFile(lines) {
-  lines.forEach((line) => {
-    console.log(line);
+const fileNames = userArguments.filter((argument) => !argument.startsWith("-"));
+let allFilesContents = [];
+
+readFiles();
+executeFlags();
+printLines();
+
+function readFiles() {
+  fileNames.forEach((fileName) => {
+    fileContent = readFile(fileName);
+    allFilesContents.push(fileContent.split("\n"));
   });
 }
 
-function executeFlag(lines) {
-  if (flag === "-n") {
-    return lines.map((line, index) => `${String(index + 1).padStart(6, " ")} ${line}`);
-  } else if (flag === "-b") {
+function readFile(fileName) {
+  const filePath = currentWorkingDirectory + fileName;
+  return fs.readFileSync(filePath, "utf8").trimEnd();
+}
+
+function executeFlags() {
+  for (const flag of flags) {
+    if (flagHandlers[flag]) {
+      allFilesContents = flagHandlers[flag]();
+    } else {
+      console.error(`cat: illegal option -- ${flag}\nusage: cat [-belnstuv] [file ...]`);
+      process.exit(1);
+    }
+  }
+}
+
+function bFlag() {
+  return allFilesContents.map((fileContent) => {
     let lineNumber = 1;
-    return lines.map((line) => {
-      if (line === "") {
+    return fileContent.map((line, index) => {
+      if (line == "") {
         return `${line}`;
       }
       return `${String(lineNumber++).padStart(6, " ")} ${line}`;
     });
-  } else {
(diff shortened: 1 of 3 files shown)


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,73 @@
...
+      console.error(`cat: illegal option -- ${flag}\nusage: cat [-belnstuv] [file ...]`);
+      process.exit(1);
+    }
+  }
+}
+
+function bFlag() {
+  return allFilesContents.map((fileContent) => {
+    let lineNumber = 1;
+    return fileContent.map((line, index) => {
```
LonMcGregor, line 52: A map is typically used to change one array into another using a function that doesn't have side effects. Here, you are changing a lineNumber variable each time the map runs. This isn't wrong, but just be careful of side effects, especially if you ever want to run things asynchronously.

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,122 @@
...
+        `ls: invalid option -- ${flag}\nusage: ls [-@ABCFGHILOPRSTUWXabcdefghiklmnopqrstuvwxy1%,] [--color=when] [-D format] [file ...]`,
+      );
+      process.exit(1);
+    }
+  }
+}
+
+function Flag1() {
+  if (printInList) {
+    outputString = outputString.replaceAll("\t", "\n");
```
LonMcGregor, line 49: I like that you are using lots of task specific functions.

But, is using global variables in this way a safe way of programming this? Is there a way that would have the data flow through your app in a more semantic way?

implement-shell-tools/ls/ls.js
```diff
@@ -0,0 +1,122 @@
...
+  for (let i = fileList.length - 1; i >= 0; i--) {
+    if (fileList[i][0] == ".") {
+      fileList.splice(i, 1);
+    }
+  }
+  return fileList;
+}
+
+function print() {
+  if (printInList == true) {
```
LonMcGregor, line 118: Think about how if conditions work, and check this line again. Are you writing code in the most optimal way?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,141 @@
...
+function cFlag() {
+  deleteOutputs();
+  allFilesData.forEach((sourceFile) => {
+    const targetFile = outputData.find((file) => file.name === sourceFile.name);
+    targetFile.byteSize = getIfTrue(sourceFile, "byteSize");
+  });
+}
+
+function deleteOutputs() {
+  if (!deleted) {
```
LonMcGregor, line 114: Can you explain your thinking that led to a design where you need to manually handle your resources like this?

LonMcGregor (commented):

You've built your own argument handling system. Is your way of handling options/arguments always going to be the best? Is there a better way of doing this?

implement-shell-tools/ls/ls.js
Alex-Jamshidi: I have rewritten many of the functions to allow the data to flow through, rather than the functions independently manipulating data stored in global variables.

implement-shell-tools/ls/ls.js
Alex-Jamshidi: ah yes, I see - updated

implement-shell-tools/wc/wc.js
Alex-Jamshidi: I think it was a little convoluted...

Given no flags displays all data, but the presence of any flags means only the flagged data is displayed.
I decided that if any flag was given: all data should be deleted and just the data for that flag added back in. 

This would be the same for all flags.

I've decided to create a separate array which lists displayed metrics from flags to simplify this.

I've also refactored a lot.

implement-shell-tools/cat/cat.js
Alex-Jamshidi: I see - I have refactored my code, so now I use a foreach to mutate the original array rather than mapping a new one.

Alex-Jamshidi:
> Good start but there is quite a bit of work still to do.
> 
> You've built your own argument handling system. Is your way of handling options/arguments always going to be the best? Is there a better way of doing this?

By which I assume you mean using the commander library? Yes this would be simpler to write and an ultimately more effective way to handle things

LonMcGregor (commented):
This is working now, and the task is complete. For the future I have a couple of suggestions to think about:
1. Yes, when you have an opportunity to use a library to do some of the "heavy lifting", like commander for argument parsing, that can be a good idea.
2. It's good that you are using task specific functions, but be aware that you don't need to do this if you think you'll only use a particular bit of code in one place. It is possible to abstract too much.

Alex-Jamshidi:
Thanks, I appreciate the feedback.

PR #637, opened 2026-07-30
London | 26-Jul-SDC | Roman Sanaye | Sprint 4 | Implement shell tool in Python

RomanSanaye:
## Learners, PR Template

Self checklist

- [x] My changes follow the style guide

## Changelist

Implemented the `wc` shell tool using Python.

The implementation supports:
- Reading command-line arguments
- Separating flags and file paths
- Counting lines, words, and characters
- Supporting `-l`, `-w`, and `-c` flags
- Handling multiple files and displaying total counts

I tested the program against the required commands:
- `wc sample-files/*`
- `wc -l sample-files/3.txt`
- `wc -w sample-files/3.txt`
- `wc -c sample-files/3.txt`
- `wc -l sample-files/*`
- `wc -w -l sample-files/3.txt`
- `wc -w -l sample-files/*`

## Questions

No questions.

Task ID: CYF-1152

### DIFF
--- a/implement-shell-tools/wc/wc.py
+++ b/implement-shell-tools/wc/wc.py
@@ -1,37 +1,39 @@
-import sys
+import argparse
 
-args = sys.argv[1:]
+# Set up command-line argument parser
+parser = argparse.ArgumentParser()
+
+# Add supported flags
+parser.add_argument("-l", action="store_true")
+parser.add_argument("-w", action="store_true")
+parser.add_argument("-c", action="store_true")
+
+# Accept one or more file paths
+parser.add_argument("paths", nargs="+")
+
+# Parse the user's command-line arguments
+args = parser.parse_args()
 
 flags = []
-paths = []
 
-# Separate flags and paths
-for arg in args:
-    if arg.startswith("-"):
-        flags.append(arg)
-    else:
-        paths.append(arg)
+# Store selected flags for the existing logic
+if args.l:
+    flags.append("-l")
+
+if args.w:
+    flags.append("-w")
+
+if args.c:
+    flags.append("-c")
+
+paths = args.paths
 
 
-total_lines = 0
-total_words = 0
-total_chars = 0
-
-
-for file in paths:
-
-    # Read file as bytes
-    with open(file, "rb") as f:
-        content = f.read()
-
-    # Count
-    lines = content.count(b"\n")
-    words = len(content.split())
-    chars = len(content)
-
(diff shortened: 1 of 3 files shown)


implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,74 @@
...
+    # Add to totals
+    total_lines += lines
+    total_words += words
+    total_chars += chars
+
+
+# Print total only when multiple files
+if len(paths) > 1:
+
+    total_output = []
```
LonMcGregor: What do you think about the output code here? Can you spot any duplication across the file?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Output for this file
+    output = []
+
+    if len(flags) == 0:
+        output = [lines, words, chars]
+
+    else:
+        if "-l" in flags:
+            output.append(lines)
```
LonMcGregor, line 42: If you run this with multiple files, how does the output look? What change could make it neater?

LonMcGregor (commented):
Good start, but some things to fix.

also, Is the choice you have used for processing arguments the most effective? Is there a library you could use to help the implementation?

RomanSanaye:
> Good start, but some things to fix.
> 
> also, Is the choice you have used for processing arguments the most effective? Is there a library you could use to help the implementation?

Hi sir @LonMcGregor 
Thanks for the feedback. I’ve changed the manual argument processing to use `argparse`.

implement-shell-tools/wc/wc.py
RomanSanaye: Thanks. I removed the duplicated output logic by creating a `get_output()` function.

implement-shell-tools/wc/wc.py
RomanSanaye: Thanks. I’ve updated the output formatting to make the results neater when using multiple files.

LonMcGregor (commented):
These argparse implementations are better, and good idea in wc to write a function to handle printing output to reduce duplication.

I have some concern that you are not using the argparse library to best effect.

In particular, if we look at the cat implementation, you use this:
```
if args.n:
    flag = "-n"
```
followed later by
```
if flag == "-n":
```

Is storing things as strings even necessary when you are using a library to help you in this way?

Similar issues exist across the other implementations.

PR #639, opened 2026-07-31
NW|26-JUI-SDC|Ahmad Hmedan|Sprint 5 |Prep exercises

AhmadHmedann:
## Learners, PR Template

Self checklist

## Changelist
DONE

Task ID: CYF-1155

### DIFF
--- a/Perp-exercises/enum8.1.py
+++ b/Perp-exercises/enum8.1.py
@@ -97,14 +97,20 @@ def read_operating_system() -> OperatingSystem:
               1.Ubuntu
               2,Arch Linux
               3.macOs""")
-    choice: int = int(input("Enter Your choice[1-3]: "))
-    while choice < 1 or choice > 3:
-        choice = int(input("Enter Your choice[1-3]: "))
+  
+    while True:
+        try:
+            choice = int(input("Enter Your choice[1-3]: "))
+            if 1 <= choice <= 3:
+              break
+        except ValueError:
+            print("please enter a valid number.")
+            
     os_map: dict[int, OperatingSystem] = {
-        1: OperatingSystem.UBUNTU,
-        2: OperatingSystem.ARCH,
-        3: OperatingSystem.MACOS,
-    }
+            1: OperatingSystem.UBUNTU,
+            2: OperatingSystem.ARCH,
+            3: OperatingSystem.MACOS,
+        }
     return os_map[choice]
 
 
--- a/Perp-exercises/enum8.1.py
+++ b/Perp-exercises/enum8.1.py
@@ -92,33 +92,51 @@ def most_available_operating_system(
     return most_available
 
 
-def main() -> None:
-    name: str = input("Name: ")
-    age: int = int(input("Age: "))
-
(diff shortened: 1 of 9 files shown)


Perp-exercises/method_exercise4.1.py
LonMcGregor, line 1: Good explanations

Perp-exercises/enum8.1.py
```diff
@@ -0,0 +1,125 @@
...
+    person1: Person = Person(name, age, preferred_operating_system)
+
+    laptops_by_operating_system = group_laptops_by_operating_system(laptops)
+    matching_laptop_count: int = how_many_match(person1, laptops_by_operating_system)
+    print("We have", matching_laptop_count, "matches")
+    most_available_os: OperatingSystem = most_available_operating_system(
+        laptops_by_operating_system
+    )
+    if person1.preferred_operating_system != most_available_os:
+        print("Are you willing to accept ", most_available_os.value, "instead")
```
LonMcGregor: If you have a question like this, what is the user expected to do? Is there some way for them to answer the question / change the outcome of the program?

Perp-exercises/enum8.1.py
```diff
@@ -0,0 +1,125 @@
...
+
+def main() -> None:
+    name: str = input("Name: ")
+    age: int = int(input("Age: "))
+
+    print("""choose an operating system:
+          1.Ubuntu
+          2,Arch Linux
+          3.macOs""")
+    choice: int = int(input("Enter Your choice: "))
```
LonMcGregor: Is there any input validation here? What is the UX like if something goes wrong?

LonMcGregor (commented):
Good work on this, I have some comments for the laptop allocation task

Perp-exercises/enum8.1.py
AhmadHmedann: Thank you for the feedback, You're right_ the original version asked the question and do not let the user to respond, so it couldn't affect the program's behaviour.

I have updated the code so the user can answer Y/N, If they accept the program will assigns them a laptop with the most available OS; otherwise, it keeps their original choice or exit if there is no suitable laptops is available

Perp-exercises/enum8.1.py
AhmadHmedann: Thanks for the feedback. I added a loop to validate the user's input. If the user enters an invalid choice, the program displays a message telling them they must enter a number between 1 and 3, then keeps asking until they enter a valid option. I think this provides a better user experience than letting the program continue with invalid input.

LonMcGregor (commented):
This is better, having an input loop is much better UX. Does your input loop properly handle all possible inputs? (hint: what if I accidentally type something totally invalid?)

AhmadHmedann:
You're right. At the moment, it only handles invalid numbers outside the valid range. If the user enters something like a letter, int() raises a ValueError. 
I think for better validation I should use try catch.

LonMcGregor (commented):
This is much better now - using a try / except to catch potential problems before they crash the program is much better UX. Task complete, good work

PR #648, opened 2026-08-04
Add helper message

LonMcGregor:
The way the tests work, it requires removing and re-adding the label. I have seen a few trainees confused about this. This adds a little note about it at the end.


test-sdc.sh
```diff
@@ -180,3 +180,5 @@ elif [[ "$1" == "implement-cowsay" ]]; then
 else
 	echo "Please run this with a valid test directory name as argument"
 fi
+
+echo "If you want to re-run the automated tests after changing some code, remove the 'Needs Review', wait a little, and then re-add the 'Needs Review' label on your PR."
```
abdishakoor-dev: Is this intended to be surfaced in the automated PR comment? If so, I think you have forgotten to append `>> testoutput.txt` at the end (similar to line 8)? As it stands it goes to stdout, so it'll only show in the Actions log.

LonMcGregor: Good catch. Updated.

test-sdc.sh
```diff
@@ -180,3 +180,5 @@ elif [[ "$1" == "implement-cowsay" ]]; then
 else
 	echo "Please run this with a valid test directory name as argument"
 fi
+
+echo "If you want to re-run the automated tests after changing some code, remove the 'Needs Review', wait a little, and then re-add the 'Needs Review' label on your PR." >> testoutput.txt
```
abdishakoor-dev: Approved but I would add the word _label_ after the first mention of 'Needs Review', to be consistent with the second mention in the same senstence.

abdishakoor-dev (approved):

PR #649, opened 2026-08-04
London | 26-SDC-July | Raihan Sharif | Sprint 4 | implement shell tools python

RaihanSharif:
## Learners, PR Template

Self checklist

## Changelist
Implemented cat, ls, wc.

Task code: CYF-1152

### DIFF
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,0 +1,14 @@
+import argparse
+
+parser = argparse.ArgumentParser(
+    prog="a simple cat implementation",
+    description="cat command line tool with the -n and -b flags"
+)
+
+parser.add_argument("-n", action="store_true", help="number all output lines")
+parser.add_argument("-b", action="store_true", help="number non-empty output lines")
+parser.add_argument("paths", nargs="+", help="file path or paths", )
+
+args = parser.parse_args();
+
+print(args)
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -10,7 +10,7 @@ parser.add_argument("-n", action="store_true", help="number all output lines")
 parser.add_argument("-b", action="store_true", help="number non-empty output lines")
 parser.add_argument("paths", nargs="+", help="file path or paths", )
 
-args = parser.parse_args();
+args = parser.parse_args()
 
 # cat returns different error messages depending on the reason the path could be read
 def read_file(path):
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -1,4 +1,5 @@
 import argparse
+import sys
 
 parser = argparse.ArgumentParser(
     prog="a simple cat implementation",
@@ -11,4 +12,25 @@ parser.add_argument("paths", nargs="+", help="file path or paths", )
 
 args = parser.parse_args();
 
-print(args)
+# cat returns different error messages depending on the reason the path could be read
+def cat_file(path):
+    try:
+        with open(path, "r",) as f:
+            content = f.read()
+    except FileNotFoundError:
+        print(f"cat: {path}: No such file or directory", file=sys.stderr)
+        return
+    except IsADirectoryError:
+        print(f"cat: {path}: Is a directory", file=sys.stderr)
+        return
+    except PermissionError:
+        print(f"cat: {path}: Permission denied", file=sys.stderr)
+        return
+
+    print(content)
+
+
+
+for path in args.paths:
+    line_num = 1
+    cat_file(path)
--- a/implement-shell-tools/cat/cat.py
+++ b/implement-shell-tools/cat/cat.py
@@ -28,11 +28,11 @@ def read_file(path):
 
 # -b (number the non-empty lines) takes priority over -n (number all lines)
 # if both are present
-def format_lines(lines, number_all=False, number_non_empty=False):
+def format_lines(lines, number_all=False, number_nonempty=False):
     """Returns a list of formatted output lines"""
     output = []
     
-    if number_non_empty:
+    if number_nonempty:
         line_num = 0
         for line in lines:
             if line == "":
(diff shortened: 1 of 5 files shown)


implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,68 @@
...
+import sys
+import os
+
+# TODO: decompose into functions to make it more modular and reusable
+
+parser = argparse.ArgumentParser(
+    prog="a simple version of wc. Takes in one or more files.",
+    description="ls command line tool which can accept -l -w -c cflags")
+
+parser.add_argument("-l", action="store_true", help="show line count", default="l")
```
SlideGauge: Each flag is declared action="store_true", default="l" (and "w", "c"). With store_true, when the flag is
absent the value falls back to that default string - and a non-empty string is truthy. So run wc -l
sample-files/3.txt: does it print only the line count, or all three? What should default be for a boolean flag so
that "flag not given" reads as false?

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,93 @@
...
+        for line in lines:
+            if line == "":
+                output.append("")
+            else:
+                line_num += 1
+                # {line_num:6} right justied number, length of at least 6
+                # {some_str:6} left justifed string, length fo at least 6
+                output.append(f"{line_num:6}\t{line}")
+    elif number_all:
+        for i, line in enumerate(lines, start=1):
```
SlideGauge, line 46: format_lines numbers with enumerate(..., start=1) and is called fresh for each file in cat_file. Run cat -n
sample-files/*.txt against the real cat - does the count reset to 1 at each file, or keep climbing? Where would the counter need to live to carry across files?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,68 @@
...
+
+args = parser.parse_args()
+
+totals = {"l": 0, "w": 0, "c": 0}
+
+for path in args.paths:
+    if (not os.path.exists(path)):
+        print(f"wc: {path}: open: No such file or directory", file=sys.stderr)
+    elif (os.path.isdir(path)):
+        print(f"wc: {path}: read: Is a directory")
```
SlideGauge: The "Is a directory" message here uses plain print (stdout), while your "No such file" on line 23 correctly
uses file=sys.stderr. Which stream should errors go to?

implement-shell-tools/wc/wc.py
```diff
@@ -0,0 +1,68 @@
...
+        print(f"wc: {path}: open: No such file or directory", file=sys.stderr)
+    elif (os.path.isdir(path)):
+        print(f"wc: {path}: read: Is a directory") 
+    elif (os.path.isfile(path)):
+        output_str = ""
+        with open(path, "r", encoding="utf-8") as file:
+            lines = file.readlines()
+
+            if (args.l):
+                if (len(lines) > 0 and lines[-1] == ""):
```
SlideGauge: lines[-1] == "" - but readlines() keeps the trailing \n, so an element is "\n", never "". Does this pop ever
actually run? Check wc -l against real wc on a file that does and doesn't end in a newline.

implement-shell-tools/ls/requirements.txt
```diff
@@ -0,0 +1 @@
+argparse
```
SlideGauge: argparse is part of the Python standard library - it doesn't need to be installed. Does this file need to
exist at all?

implement-shell-tools/wc/requirements.txt
```diff
@@ -0,0 +1 @@
+argparse
```
SlideGauge: argparse is part of the Python standard library - it doesn't need to be installed. Does this file need to
exist at all?

implement-shell-tools/cat/cat.py
```diff
@@ -0,0 +1,93 @@
...
+)
+
+parser.add_argument("-n", action="store_true", help="number all output lines")
+parser.add_argument("-b", action="store_true", help="number non-empty output lines")
+parser.add_argument("paths", nargs="+", help="file path or paths", )
+
+args = parser.parse_args()
+
+# cat returns different error messages depending on the reason the path could be read
+def read_file(path):
```
SlideGauge, line 16: Really nice - distinguishing FileNotFoundError / IsADirectoryError / PermissionError with matching messages,
returning (content, error) and printing to stderr with a proper exit code.

SlideGauge (commented):
Good job, I left several comments, could you fix them please?

implement-shell-tools/cat/cat.py
RaihanSharif: yes, the numbering starts from 1 for each file in the cat input
```
cat -n sample-files/*
     1  Once upon a time...
     1  It looked delicious.
     4
```
python cat:
```
python3 cat.py -n sample-files/*
     1  Once upon a time...
     1  It looked delicious.
     4
```
You asked this in the nodeJS version of cat as well. Are you hinting at something, and I'm just not getting it?

implement-shell-tools/ls/requirements.txt
RaihanSharif: I didn't realise. Made an assumption.

implement-shell-tools/wc/requirements.txt
RaihanSharif: Nope.

implement-shell-tools/wc/wc.py
RaihanSharif: This was a typo. Fixed.

implement-shell-tools/wc/wc.py
RaihanSharif: ...uh, yeah, that's bad. 

I've now added a check that if no flags are supplied, then just set them all to true.
Which then means that wc prints the lines, word count, and bytes of the files.

Otherwise, the flags are processed individually, so that any combination of the three will work.

implement-shell-tools/wc/wc.py
RaihanSharif: Thank you for pointing that out, I 'translated' my js code too literally without accounting for how readlines() or how wc actually counts lines. 

wc counts the number of '\n' characters, not necessarily non-empty lines. Last empty line is excluded because it doesn't have a '\n' in it.

I've changed my implementation to just read the file, and count the number of '\n' characters. The output now matches the real wc.

It also made my -w logic a bit simpler.

PR #652, opened 2026-08-07
London | 6-SDC-Aug | Boshra Mahmoudi | Sprint 3 |  Implement shell tools

BoshraM:

This pull request is related to the Shell pipelines exercise task in the CYF Circulum sprint2.

Task ID: CYF-1150


implement-shell-tools/cat/cat.js
```diff
@@ -0,0 +1,48 @@
+import { readFile } from "node:fs/promises";
+import process from "node:process";
+
+const argv = process.argv.slice(2);
```
LonMcGregor: Is there a library or API you could use to handle arguments for you?

implement-shell-tools/wc/wc.js
```diff
@@ -0,0 +1,78 @@
...
+      process.stdout.write(`       ${lines}      ${words}      ${bytes} ${file}\n`);
+    }
+      
+  } catch(err) {
+    console.error(err.message);
+  }
+}
+
+if (files.length > 1 ) {
+  if (countLines && countWords) {
```
LonMcGregor: There's some repeated patterns of code here, can you find a way to reduce it?

LonMcGregor (commented):
Good start, but I think there are some changes that could be made

implement-shell-tools/cat/cat.js
BoshraM: Hi @LonMcGregor , thanks for raising this. I’ve updated the code to use `import { parseArgs } from "node:util";`.

implement-shell-tools/wc/wc.js
BoshraM: @LonMcGregor , I’ve created a function that can be reused in the main logic, so I don’t need to repeat the same code.

LonMcGregor (commented):
Good work. Using a library for argument handling is a good idea as it can be fiddly and difficult to maintain otherwise. Good work refactoring the repeated print pattern

